PR #29, opened 2017-09-27
Remove Boneyard

jinlccs:


jinlccs:
Remove Boneyard and Update document.

PR #56, opened 2017-11-01
Fetch kubernetes so that version is correctly set as version uses "gi… …t describe --tags"

sanjeevm0:
…s argument


hongzhili:
It should be fine with Ubuntu deployment. But it may not work for CoreOS deployment (nodes use reporting service to find API server).

hongzhili:
@jinlccs

PR #140, opened 2018-01-26
Enable custom metric api server by default

resouer:
1. Add all needed yamls for custom metric api
2. Add rolebindings and roles
3. Add sleep tag in service order file


resouer:
ping @hongzhili

PR #166, opened 2018-02-26
Autoscaling az

resouer:
Should be able to support this feature.


src/ClusterBootstrap/template/kubeconfig.yaml.template
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
```
hongzhili: Is this a copy of "src/ClusterBootstrap/template/kubelet/worker-kubeconfig.yaml.template" ?

resouer: Not really, we recently changed cluster scaler to local usage only so the cert path will be different

hongzhili: can you create a sub-folder under src/ClusterBootstrap/template and put this file to the sub-folder?

resouer: yes!

resouer: Fixed! Thanks!

PR #254, opened 2019-04-19
add watchdog/job-exporter

xudifsd:
Working in progress.

TODOs:

- [x] Current hard code some IP and config, should change it later.
- [x] Currently using xudifsd/job-exporter etc. Should change it to openpai/job-exporter:dlts.


xudifsd:
Still hard coded prometheus IP:PORT in grafana config.

src/ClusterBootstrap/services/monitor/grafana-config.yaml
```diff
@@ -0,0 +1,157 @@
+{% set prometheus_url = "192.168.255.1:9091" %}
```
hongzhili: How about "{{ cnf["api-server-ip"] }}:9091" ?

xudifsd: This assumes api-server is always co-allocated with prometheus, we can assume that?

xudifsd:
@hongzhili I just saw there are many dashboards available in old Grafana, some of them are still in use, for example `GPU usage` from `Cluster Status`, do I need to import them? If we import them, we will need to keep old influex & collected.

src/ClusterBootstrap/services/monitor/grafana-config.yaml
xudifsd: Have tried this, the value will be `10.3.0.1` in dltsp40 cluster, the grafana pod can not connect to this address, not sure what this IP is.

xudifsd:
There will be two dashboard, one:

image

another:
image

There are several problems:
* Current available gpu is get by watchdog requesting k8s's node api, I submitted a job requiring 1 gpu, and job is running, but available GPU from k8s's node api doesn't change, I'm not sure why.
* Original `Cluster Status` have two more columns: `Reserved GPU` and `Total Jobs`, not sure how to get these two numbers.

xudifsd:
From k8s issues https://github.com/kubernetes/kubernetes/issues/43045 and [doc](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/node-allocatable.md), the allocatable is defined as:

```
[Allocatable] = [Node Capacity] - [Kube-Reserved] - [System-Reserved] - [Hard-Eviction-Threshold]
```

It's not available resources. I will find another way to do this.

src/ClusterBootstrap/services/monitor/pre-render.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+dir=`dirname $0`
+
+file_name=${dir}/grafana-config.yaml
+
+# create configmap
+for i in `find ${dir}/grafana-config/ -type f -regex ".*json" ` ; do
+    echo --from-file=$i
+done | xargs ${dir}/../../deploy/bin/kubectl --namespace=kube-system create configmap grafana-configuration --dry-run -o yaml >> $file_name
```
hongzhili: call kubectl directly doesn't work. please use ./deploy.py kubectl ***

hongzhili:
How do I launch the new monitor system? still ./deploy.py kubernetes start monitor?

src/ClusterBootstrap/services/monitor/pre-render.sh
xudifsd: This is used to generate configmap yaml file through files, it doesn't need to be called via `deploy.py kubectl`.

xudifsd:
@hongzhili yes.

hongzhili:
Should I start to merge this branch? does it still WIP?  @xudifsd

xudifsd:
Possible problem: grafana still use dns to locate prometheus service. If dns has problems grafana will broken.

We can merge this and change it in future commits.

xudifsd:
@hongzhili I have resolved the conflicts, and find a way to get static IP of prometheus.

You can merge this PR now.

src/ClusterBootstrap/deploy.py
```diff
@@ -261,6 +261,8 @@ def update_config():
         config["elasticsearch_node"] = config["webportal_node"]
     if ("mysql_node" not in config):
         config["mysql_node"] = None if len(get_node_lists_for_service("mysql"))==0 else get_node_lists_for_service("mysql")[0]
+    if ("host" not in config["prometheus"]):
+        config["prometheus"]["host"] = None if len(get_node_lists_for_service("prometheus"))==0 else get_node_lists_for_service("prometheus")[0]
```
xudifsd: Just deployed in p40 bed, this config will become a public IP. How can I get a private IP?

PR #256, opened 2019-04-24
Mount all IB devices for regular/distributed jobs

Anbang-Hu:


hongzhili:
will not merge since this is not a generic change.

PR #260, opened 2019-05-04
handling user uid/gid if winbind not connected to RestfulAPI

deepak-ms:


deepak-ms:
Pending comprehensive tests. Will update once those completed.

src/RestAPI/dlwsrestapi.py
```diff
@@ -597,12 +597,29 @@ def get(self):
...
+        parser.add_argument('groups')
         args = parser.parse_args()
-        username = args["userName"]
-        userId = args["userId"]
+
         ret = {}
-        ret["status"] = JobRestAPIUtils.AddUser(username,userId)
+        userName = args["userName"]
+        if args["uid"] is None or len(args["uid"].strip()) == 0:
+            uid = 0
```
hongzhili: uid & gid cannot be 0. It becomes root...

deepak-ms: Fixed.

PR #269, opened 2019-05-16
Ssh feature

hao1939:
1. Make the create endpoints API idempotent, use the same id to tolerate duplicate requests.
2. Stop endpoints after the job is not running.
3. Support juptyer notebook port.


hao1939:
Hi @hongzhili , after you have done the reviewing, we can merge it first. Will check-in 'set user in container' with another PR.

PR #278, opened 2019-05-27
Distribute job support in multiple physical clusters

Gerhut:


src/WebUI/dotnet/WebPortal/Views/Home/JobSubmission.cshtml
```diff
@@ -613,14 +620,14 @@
 
                         <md-input-container flex="50">
                             <label> Total number of GPUs </label>
-                            <input ng-value="current.numpsworker * 8" disabled type="number" md-no-asterisk>
+                            <input ng-value="current.numpsworker * gpus[extras.gpuType]" disabled type="number" md-no-asterisk>
```
hongzhili: why gpuType? is it the number of GPUs per machine?

Gerhut: `extras.gpuType` is the current GPU type user selected, and the `gpus` scope variable is a type-count map, so `gpus[extra.gpuType]` is the count of the user selected GPU type.

hongzhili: got it. Thanks.

PR #283, opened 2019-05-28
add alert manager

xudifsd:
we can deploy with following config to enable alert-manager:

``` yaml
alert-manager:
  configured: True
  smtp_url: smtp.office365.com:587
  smtp_from: xxx@microsoft.com
  smtp_auth_username: xxx@microsoft.com
  smtp_auth_password: xxx
  receiver: yyy@example.com
```

auto-healing can be enabled via [this method](https://static.sched.com/hosted_files/kccncna17/f1/slides.pdf), i.e. leverage webhook of alert-manager to trigger some external process to do something on alerts.


xudifsd:
alert email will looks like:

image

I do not know how to get external IP for master node.

PR #284, opened 2019-05-28
Fix ssh config, user permission and etc

hao1939:
Some fixes:

1. Fix ~/.ssh/config override problem, now multiple dist job works.
1. Fix ~/.ssh permission issues.
1. Run jupyter as normal user (instead of root).
1. Add a column "lastUpdated" for jobs

### DIFF
--- a/src/utils/MySQLDataHandler.py
+++ b/src/utils/MySQLDataHandler.py
@@ -91,6 +91,7 @@ class DataHandler:
                     `jobMeta` LONGTEXT  NULL,
                     `jobLog` LONGTEXT  NULL,
                     `retries`             int    NULL DEFAULT 0,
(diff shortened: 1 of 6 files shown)


src/utils/MySQLDataHandler.py
```diff
@@ -91,6 +91,7 @@ def CreateTable(self):
                     `jobMeta` LONGTEXT  NULL,
                     `jobLog` LONGTEXT  NULL,
                     `retries`             int    NULL DEFAULT 0,
+                    `lastUpdated` DATETIME     DEFAULT CURRENT_TIMESTAMP NOT NULL,
```
hongzhili: Please make sure this change won't cause backward compatibility issue.

PR #287, opened 2019-05-29
add job status and gpu usage dashboard

xudifsd:
Add job status and gpu usage dashboard.

Job status dashboard looks like:

image

Job uses little resource because it's `sleep infinity` job.

Gpu usage dashboard looks like:

image


xudifsd:
@Gerhut job status url will be `http://$IP:3000/dashboard/db/job-status?var-job_name=$job_id` where $job_id is job id like `27cd4600-692e-4410-b514-a2547d88ad71`. And gpu usage url will be `http://$IP:3000/dashboard/db/gpu-usage`.

PR #330, opened 2019-06-07
disable first end mapped point display

LeoHongyi:


src/WebUI/dotnet/WebPortal/Views/Home/JobDetail.cshtml
```diff
@@ -182,10 +182,20 @@
...
 
             var endpointsStr = $.map(endpoints, function (endpoint) {
+                //console.log(endpoint.podName);
+                var task = "";
+
+                if (endpoint.podName) {
+                    task = endpoint.podName.split("-").pop();
+                }
+
+                if (task.indexOf("ps0") !== -1) {
```
hao1939: Why should we ignore the endpoints on "ps0"?
The "ipython" port would be hosted on "ps0".

PR #366, opened 2019-06-19
add DLWS_USER_EMAIL env to pod, and add alert rule for idle jobs

xudifsd:
Add DLWS_USER_EMAIL env to job template. Current assume users are all from corp.

Also add alert rule for idle jobs.

### DIFF
--- a/src/ClusterBootstrap/services/monitor/alert-manager.yaml
+++ b/src/ClusterBootstrap/services/monitor/alert-manager.yaml
@@ -84,22 +84,29 @@ data:
       routes:
       - receiver: task_user
         repeat_interval: 4h
-        group_by: [alertname, user_email]
+        group_by: [alertname, user_email, cluster]
         match_re:
           type: user_alert
+          alertname: "zero-gpu-usage"
     receivers:
     - name: "alert-email"
       email_configs:
         - to: {{ alert_info["receiver"] }}
           html: '{{ "{{" }} template "email.html" . {{ "}}" }}'
           headers:
-            subject: '{{ cnf["cluster_name"] }}: {{ "{{" }} template "__subject" . {{ "}}" }}'
+            subject: '{{ "{{" }} .GroupLabels.cluster {{ "}}" }}: {{ "{{" }} template "__subject" . {{ "}}" }}'
     - name: "task_user"
       email_configs:
+        {% if cnf["alert-manager"]["alert_users"] %}
+        - to: '{{ "{{" }} .GroupLabels.user_email {{ "}}" }},{{ alert_info["receiver"] }}'
+        {% else %}
         - to: '{{ alert_info["receiver"] }}'
-          html: '{{ "{{" }} template "user_alert.html" . {{ "}}" }}'
+        {% endif %}
+          html: '{{ "{{" }} template "idle_gpu.html" . {{ "}}" }}'
           headers:
-            #To: '{{ "{{" }} .GroupLabels.user_email {{ "}}" }}'
-            #CC: '{{ alert_info["receiver"] }}'
-            subject: '{{ cnf["cluster_name"] }}: {{ "{{" }} template "__subject" . {{ "}}" }}'
+            {% if cnf["alert-manager"]["alert_users"] %}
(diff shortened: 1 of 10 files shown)


src/ClusterManager/job_manager.py
```diff
@@ -169,6 +169,8 @@ def SubmitRegularJob(job):
             jobParams["mountpoints"].append(mp)
 
         userAlias = getAlias(jobParams["userName"])
+        userEmail = userAlias + "@microsoft.com" # assume users are corp users
```
deepak-ms: I think jobParams["userName"] contains full emailId. If that's the case, then we can use it directly instead of hardcoding microsoft.com.

src/ClusterBootstrap/services/monitor/alert-manager.yaml
```diff
@@ -80,11 +80,22 @@ data:
...
     receivers:
     - name: "alert-email"
       email_configs:
         - to: {{ alert_info["receiver"] }}
           html: '{{ "{{" }} template "email.html" . {{ "}}" }}'
           headers:
             subject: '{{ cnf["cluster_name"] }}: {{ "{{" }} template "__subject" . {{ "}}" }}'
+    - name: "task_user"
+      email_configs:
+        - to: '{{ "{{" }} .GroupLabels.user_email {{ "}}" }}'
```
deepak-ms: Will there be any of our team DL in 'cc'?

src/ClusterManager/job_manager.py
xudifsd: Ok, I'll use that then.

src/ClusterBootstrap/services/monitor/alert-manager.yaml
xudifsd: Good suggestion.

xudifsd:
> Killing job after 8 hours is not included in this change. right?

Yes, this PR only for email.

src/ClusterBootstrap/services/monitor/alerting/jobs.rules
```diff
@@ -0,0 +1,10 @@
+groups:
+    - name: idle-job
+      rules:
+      - alert: low-gpu-usage
+        expr: avg(task_gpu_percent) by (pod_name, instance, user_email) == 0
+        for: 4h
+        labels:
+          type: user_alert
+        annotations:
+          summary: "Your job {{$labels.pod_name}} in {{$labels.instance}} have not used gpu for 4 hours. Please kill the job if you do not need it anymore"
```
Anbang-Hu: 1. Is it possible to add a link to the Job?
2. For distributed jobs, will each idle pod trigger an email? If so, is it possible to aggregate them at job level?

src/ClusterBootstrap/services/monitor/alerting/jobs.rules
```diff
@@ -0,0 +1,10 @@
+groups:
+    - name: idle-job
+      rules:
+      - alert: low-gpu-usage
```
Anbang-Hu: zero

src/ClusterBootstrap/services/monitor/alerting/jobs.rules
Anbang-Hu: I see that it's already assembled into one email. But still can we make them cleaner by grouping them into jobs?

xudifsd: 1. Yes, we can add a link, but to do so, alert manager must know webportal's url, maybe hub's, how can we get it? Through config?
2. Yes, in that case, we avg all task role's gpu usage and still check if it is 0?

src/ClusterBootstrap/services/monitor/alerting/jobs.rules
xudifsd: sure

src/ClusterBootstrap/services/monitor/alerting/jobs.rules
xudifsd: Also for 1, where the link pointing to? Can we assume all jobs are from hub instead of cluster local webportal?

xudifsd:
Current will send to admin for now

PR #372, opened 2019-06-20
Feature: interactive port and tensorboard

hao1939:
Features:
1. Interactive ports [DONE]
1. Tensorboard [DONE]

Design restriction:
1. Tensorboard would working with `--logdir=/job/logs`
1. All interactive  (and tensorboard) ports serve on ‘master’ role.
1. Listen on the port in range 40000-49999. (Try to avoid port overlay with NodePort.)
1. Interactive port name should less than 16 characters. (We use port name as part of the k8s service name, k8s have length limitation on the service name.)

API payload example for `create` endpoints:
`{ "jobId": "bd3d090a-53b6-4616-9b6c-fe4a86fd68ea", "endpoints": ["ssh", "ipython", "tensorboard", {"name": "port1", "podPort": "49999"}]}
`

### DIFF
--- a/src/RestAPI/dlwsrestapi.py
+++ b/src/RestAPI/dlwsrestapi.py
@@ -1007,6 +1007,7 @@ class Endpoint(Resource):
 
         for [_, endpoint] in endpoints.items():
             ret = {
+                "id": endpoint["id"],
                 "name": endpoint["name"],
                 "username": endpoint["username"],
                 "status": endpoint["status"],
@@ -1014,6 +1015,8 @@ class Endpoint(Resource):
                 "podName": endpoint["podName"],
                 "domain": config["domain"],
             }
+            if "podPort" in endpoint:
+                ret["podPort"] = endpoint["podPort"]
             if endpoint["status"] == "running":
                 if endpoint["hostNetwork"]:
                     port = int(endpoint["port"])
@@ -1050,9 +1053,15 @@ class Endpoint(Resource):
                 for i in range(nums[role]):
                     pod_names.append(job_id + "-" + role + str(i))
 
-        # endpoints should be in ["ssh", "ipython"]
-        if any(elem not in ["ssh", "ipython"] for elem in requested_endpoints):
-            return ("Bad request, endpoints only allowed in [\"ssh\", \"ipython\"]: %s" % requested_endpoints), 400
+        interactive_ports = []
+        # endpoints should be ["ssh", "ipython", "tensorboard", {"name": "port name", "podPort": "port on pod in 50000-59999"}]
+        for interactive_port in [ elem for elem in requested_endpoints if elem not in ["ssh", "ipython", "tensorboard"] ]:
+            if any(required_field not in interactive_port for required_field in ["name", "podPort"]):
+               # if ["name", "port"] not in interactive_port:
+               return ("Bad request, interactive port should have \"name\" and \"podPort\"]: %s" % requested_endpoints), 400
+            if int(interactive_port["podPort"]) < 50000 or int(interactive_port["podPort"]) > 59999:
+               return ("Bad request, interactive podPort should in range 50000-59999: %s" % requested_endpoints), 400
+            interactive_ports.append(interactive_port)
 
         # HostNetwork
         if "hostNetwork" in job_params and job_params["hostNetwork"] == True:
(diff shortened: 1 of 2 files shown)


src/RestAPI/dlwsrestapi.py
```diff
@@ -1078,7 +1089,7 @@ def endpoint_exist(endpoint_id):
         if "ssh" in requested_endpoints:
             # setup ssh for each pod
             for pod_name in pod_names:
-                endpoint_id = "endpoint-" + pod_name + "-ssh"
+                endpoint_id = "e-" + pod_name + "-ssh"
```
Anbang-Hu: Why do we want to do this? This is not as descriptive as "endpoint".

src/RestAPI/dlwsrestapi.py
```diff
@@ -1050,9 +1053,17 @@ def post(self):
                 for i in range(nums[role]):
                     pod_names.append(job_id + "-" + role + str(i))
 
-        # endpoints should be in ["ssh", "ipython"]
-        if any(elem not in ["ssh", "ipython"] for elem in requested_endpoints):
-            return ("Bad request, endpoints only allowed in [\"ssh\", \"ipython\"]: %s" % requested_endpoints), 400
+        interactive_ports = []
+        # endpoints should be ["ssh", "ipython", "tensorboard", {"name": "port name", "podPort": "port on pod in 40000-49999"}]
```
Anbang-Hu: Using the same range as ssh_port can cause collision.

src/RestAPI/dlwsrestapi.py
```diff
@@ -1050,9 +1053,17 @@ def post(self):
...
+        for interactive_port in [ elem for elem in requested_endpoints if elem not in ["ssh", "ipython", "tensorboard"] ]:
+            if any(required_field not in interactive_port for required_field in ["name", "podPort"]):
+                # if ["name", "port"] not in interactive_port:
+                return ("Bad request, interactive port should have \"name\" and \"podPort\"]: %s" % requested_endpoints), 400
+            if int(interactive_port["podPort"]) < 40000 or int(interactive_port["podPort"]) > 49999:
```
Anbang-Hu: Extract this into a variable

src/RestAPI/dlwsrestapi.py
```diff
@@ -1050,9 +1053,17 @@ def post(self):
...
+                return ("Bad request, interactive podPort should in range 40000-49999: %s" % requested_endpoints), 400
+            if len(interactive_port["name"]) > 16:
```
Anbang-Hu: Why do we have this restriction? Where is this "name" coming from?

src/RestAPI/dlwsrestapi.py
```diff
@@ -1123,6 +1134,60 @@ def endpoint_exist(endpoint_id):
             else:
                 print("Endpoint {} exists. Skip.".format(endpoint_id))
 
+        # Only open tensorboard on the master
+        if 'tensorboard' in requested_endpoints:
+            if job_type == "RegularJob":
+                pod_name = pod_names[0]
+            else:
+                # For a distributed job, we set up jupyter on first worker node.
```
Anbang-Hu: Change it to make the comment accurately describe the situation.

Anbang-Hu: If it's reused so often, let's make it a function.

src/RestAPI/dlwsrestapi.py
hao1939: To reduce the length of `id`, k8s have a limit of 63 characters.

src/RestAPI/dlwsrestapi.py
hao1939: 1. There would be no problem for k8s newtork.
2. For HostNetwork, the range 40000-49999 is for ssh/ipython/tensorboard. When use the HostNetwork, we assume that one pod would take the whole machine, so, the collision is controllable.

src/RestAPI/dlwsrestapi.py
hao1939: Comes from that k8s have a length limitation on 'id'.

src/RestAPI/dlwsrestapi.py
```diff
@@ -1123,6 +1134,60 @@ def endpoint_exist(endpoint_id):
             else:
                 print("Endpoint {} exists. Skip.".format(endpoint_id))
 
+        # Only open tensorboard on the master
+        if 'tensorboard' in requested_endpoints:
```
Anbang-Hu: Use double quotes for consistency everywhere

PR #378, opened 2019-06-24
Job manager should not crash due to temporary failure in SQL server name resolution

Anbang-Hu:
…ion for sql server


Anbang-Hu:
@hao1939 In `cleanup_endpoints` method, `status` variable is not assigned below
```
dead_endpoint["status"] = status
```
It will crash the program if unfortunately this code path is hit. What should be the value for `status`?

PR #383, opened 2019-06-25
Add command validation & fix interactive port display.

Gerhut:


src/WebUI/dotnet/WebPortal/Controllers/HomeController.cs
```diff
@@ -926,10 +926,14 @@ public IActionResult JobDetail()
                 return RedirectToAction("Index", "Home");
             }
             var cluster = HttpContext.Request.Query["cluster"];
+            if (!Startup.Clusters.ContainsKey("cluster"))
+            {
+                return RedirectToAction("Index", "Home");
```
Anbang-Hu: This change seems to cause redirection to Home page when clicking on any Job Id.

PR #407, opened 2019-07-09
add support for notify user about job state change

xudifsd:
Current only notify users when their jobs failed.

Will need to add following config to jobmanager:
```yaml
job-manager:
  notifier:
    cluster: $user_friendly_cluster_name # e.g. Azure-EastUS-V100
    alert-manager-url: $alert_manager_url # e.g. http://localhost:9093/alert-manager
```

This will use alert-manager to send email.

Or add following config to jobmanager to use smtp to send out email directly, if alert-manager was not deployed:
```yaml
job-manager:
  notifier:
    smtp-url: smtp.office365.com:587
    smtp-from: xxx@example.com
    smtp-auth-username: xxx@example.com
    smtp-auth-password: xxxx
```

### DIFF
--- a/src/ClusterManager/job_manager.py
+++ b/src/ClusterManager/job_manager.py
@@ -180,7 +180,7 @@ def AutoApproveJob(job):
 
 UnusualJobs = {}
 
-def UpdateJobStatus(job, notifier):
+def UpdateJobStatus(job, notifier=None):
     dataHandler = DataHandler()
     jobParams = json.loads(base64.b64decode(job["jobParams"]))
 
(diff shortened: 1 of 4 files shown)


src/ClusterManager/job_manager.py
```diff
@@ -618,7 +619,7 @@ def AutoApproveJob(job):
 
 UnusualJobs = {}
 
-def UpdateJobStatus(job):
+def UpdateJobStatus(job, notifier):
```
hongzhili: should we add the default value to notifier for backwards compatibility. 
def UpdateJobStatus(job, notifier = None):
......
if notifier is not None:
   XXX

PR #410, opened 2019-07-11
add cluster gpu statistic dashboard

xudifsd:
add a cluster gpu statistic dashboard for admin to view.

image


hongzhili:
Great! thanks.

PR #414, opened 2019-07-12
Fix issue of not fetching through restfulapi without login

LeoHongyi:


LeoHongyi:
@hongzhili can you please check

PR #425, opened 2019-07-19
Fix the issue of using other user credentials

LeoHongyi:


src/WebUI/dotnet/WebPortal/Controllers/dlwsController.cs
```diff
@@ -145,6 +145,7 @@ public async Task<string> GetLog(string jobId)
         private async Task<Tuple<bool, string>> processRestfulAPICommon()
         {
             var passwdLogin = false;
+            var loginedEmail = HttpContext.Session.GetString("Email");
```
Anbang-Hu: loginEmail

PR #428, opened 2019-07-23
Fix the issue of using other user credentials

LeoHongyi:


src/WebUI/dotnet/WebPortal/Controllers/dlwsController.cs
```diff
@@ -301,7 +315,7 @@ public async Task<ActionResult> Get(string op)
                                    System.Text.Encodings.Web.UrlEncoder.Default.Encode(item.Value) + "&";
                         }
                     }
-                    url += "userName=" + HttpContext.Session.GetString("Email") + "&";
+                    url += "userName=" + username + "&";
                     url += "userId=" + HttpContext.Session.GetString("uid") + "&";
```
Gerhut: How about these session fields when password authentication is used?

PR #460, opened 2019-08-08
upgrade to k8s 1.15.2

xudifsd:
add following command to deploy.py

* `upgrade_masters` Upgrade the master nodes.
* `upgrade_workers [nodes]` Upgrade the worker nodes. If no additional node is specified, all nodes will be updated.
* `upgrade [nodes]` Upgrade the cluster and nodes. If no additional node is specified, all nodes will be updated.

To upgrade a cluster, a command of `./deploy.py upgrade` should be enough. The whole upgrading process do not stop the docker daemon and containers, but my experiment shows it will restart container, notification for customer will be needed.


src/ClusterBootstrap/storage/glusterFS/gk-deploy
```diff
@@ -207,7 +207,7 @@ check_pods() {
       break
     fi
     sleep 2
-    pods=$(${CLI} get pod --no-headers --show-all --selector="${1}" 2>&1)
+    pods=$(${CLI} get pod --no-headers --selector="${1}" 2>&1)
```
xudifsd: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.10.md#deprecations

> --show-all (which only affected pods and only for human readable/non-API printers) is now defaulted to true and deprecated. It will be inert in 1.11 and removed in a future release.

src/ClusterBootstrap/template/kubelet/kubelet.service.template
```diff
@@ -22,17 +22,14 @@ ExecStartPre=/bin/bash -c 'if  lspci | grep -qE "[0-9a-fA-F][0-9a-fA-F]:[0-9a-fA
...
   --enable-server=true \
   --register-node=true \
-  --feature-gates="Accelerators=true,DevicePlugins=true" \
-  --allow-privileged=true \
+  --feature-gates="DevicePlugins=true,PodShareProcessNamespace=true" \
   --pod-manifest-path=/etc/kubernetes/manifests \
   --pod-infra-container-image={{cnf["dockers"]["container"]["podinfra"]["fullname"]}} \
   --network-plugin=cni \
   --cluster_dns={{cnf["dns-server-ip"]}} \
-  --docker-disable-shared-pid \
```
xudifsd: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.10.md#deprecations

> The kubelet --docker-disable-shared-pid flag, which runs docker containers with a process namespace that is shared between all containers in a pod, is now deprecated and will be removed in a future release. It is replaced by v1.Pod.Spec.ShareProcessNamespace, which configures this behavior. This field is alpha and can be enabled with --feature-gates=PodShareProcessNamespace=true.

src/ClusterBootstrap/template/kubelet/kubelet.service.template
```diff
@@ -22,17 +22,14 @@ ExecStartPre=/bin/bash -c 'if  lspci | grep -qE "[0-9a-fA-F][0-9a-fA-F]:[0-9a-fA
 # 
 # https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/
 ExecStart=/opt/bin/kubelet \
-  --require-kubeconfig=true \
   --container-runtime={{'remote' if cnf["kube_custom_cri"] else 'docker'}} \
   --enable-server=true \
   --register-node=true \
-  --feature-gates="Accelerators=true,DevicePlugins=true" \
-  --allow-privileged=true \
```
xudifsd: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.11.md#before-upgrading

> The Kubelet's deprecated --allow-privileged flag now defaults to true. This enables users to stop setting --allow-privileged in order to transition to PodSecurityPolicy. Previously, users had to continue setting --allow-privileged, because the default was false.

PR #464, opened 2019-08-19
add per vc gpu usage dashboard

xudifsd:
Add per vc gpu usage dashboard.

Webportal should change url to `http://$url/dashboard/db/per-vc-gpu-statistic?var-vc_name=$vc_name` where $vc_name should be actual vc name like `vc1` or `platform`.

image


xudifsd:
Remember to redeploy watchdog and add config `watchdog.vc_url`, e.g. `http://$restapi_url/ListVCs?userName=$admin_user`.

PR #479, opened 2019-09-03
regular cluster automatic deployment on GPU machine

YinYangOfDao:


src/ClusterBootstrap/deploy.py
```diff
@@ -2755,14 +2780,11 @@ def kubernetes_label_nodes( verb, servicelists, force ):
...
-# Label kubernete nodes with gpu types.
+# Label kubernete nodes with gpu types.skip for CPU workers
 def kubernetes_label_GpuTypes():
-    nodes = get_nodes(config["clusterId"])
-    gpuType_config = fetch_config(config, ["gpuType_config"])
-    for gpuType, nodes in gpuType_config:
-        for node in nodes:
-            nodename = kubernetes_get_node_name(node)
-            kubernetes_label_node("--overwrite", nodename, "gpuType="+gpuType)
+    for nodename,nodeInfo in config["machines"].items():
```
hongzhili: What's the default value of gpu-type? What if gpu-type is not set for a node?

hongzhili: Also, does this work with on-perm machine deployment?

src/ClusterBootstrap/deploy.py
```diff
@@ -2810,6 +2832,10 @@ def start_one_kube_service(fname):
         except Exception as e:
             pass
 
+    if fname == "./deploy/services/jobmanager/jobmanager.yaml":
```
hongzhili: Is there any better to create the configman with jobmanager? consider this: if we change the name or path of jobmanager.yaml later, this hard-coded path will be invalided. also when we stop jobmanager, we need to make sure that this configmap should be deleted as well.

src/ClusterBootstrap/deploy.py
```diff
@@ -2908,6 +2934,23 @@ def run_docker_image( imagename, native = False, sudo = False ):
         else:
             run_docker( matches[0], prompt = imagename, dockerConfig = dockerConfig, sudo = sudo )
 
+def gen_dns_config_script():
```
hongzhili: This is a good idea. But please leverage jinja2 templates to create the dns.sh. Also, consider the case for on-perm as well.

src/ClusterBootstrap/deploy.py
```diff
@@ -2908,6 +2934,23 @@ def run_docker_image( imagename, native = False, sudo = False ):
...
+    dnsf = open("./scripts/dns.sh","w")
+    dnsf.write("sudo systemctl disable systemd-resolved.service\nsudo systemctl stop systemd-resolved\necho \"dns=default\" | sudo tee -a /etc/NetworkManager/NetworkManager.conf")
+    dnsf.write("sudo rm /etc/resolv.conf\necho \"nameserver 8.8.8.8\" | sudo tee -a /etc/resolv.conf\n")
+    dnsf.write("echo \"search {}.cloudapp.azure.com\" | sudo tee -a /etc/resolv.conf\n".format(config["azure_cluster"][config["cluster_name"]]["azure_location"]))
+    dnsf.write("sudo chattr -e /etc/resolv.conf\nsudo chattr +i /etc/resolv.conf\n")
+    dnsf.close()
+
+def gen_pass_secret_script():
```
hongzhili: The same here, please consider to use jinja2 template.

src/ClusterBootstrap/deploy.py
```diff
@@ -3577,6 +3628,17 @@ def run_command( args, command, nargs, parser ):
             print "Error: kubernetes need a subcommand."
             exit()
 
+    elif command == "gpulabel":
+        kubernetes_label_GpuTypes()
+
+    elif command == "gen_scripts":
```
hongzhili: current runscript function should render the scripts file with jinja2, you can skip the generation process.

src/ClusterBootstrap/deploy.py
```diff
@@ -3577,6 +3628,17 @@ def run_command( args, command, nargs, parser ):
...
+        # print(config["azure_cluster"].keys())
+        gen_dns_config_script()
+        gen_pass_secret_script()
+
+    elif command == "setconfigmap":
```
hongzhili: This may not work. we cannot assume kubectl is working on dev machine.

src/ClusterBootstrap/install_prerequisites.sh
```diff
@@ -1,42 +1,42 @@
...
+        curl \
+        python-dev \
+        python-pip \
+        wget \
+        cpio \
+        mkisofs \
+        apt-transport-https \
+        openssh-client \
+        ca-certificates \
+        network-manager
```
hongzhili: why do we need network-manager?

src/ClusterBootstrap/install_prerequisites.sh
```diff
@@ -1,42 +1,42 @@
...
+sudo apt-key --keyring /etc/apt/trusted.gpg.d/Microsoft.gpg adv \
+     --keyserver packages.microsoft.com \
+     --recv-keys BC528686B50D79E339D3721CEB3E94ADBE1229CF
+
+sudo apt-get update
+sudo apt-get install -y --no-install-recommends
+curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash 
+
+# pip install --upgrade pip
+sudo pip install setuptools && pip install pyyaml && pip install jinja2 && pip install requests && pip install tzlocal && pip install pycurl
```
hongzhili: sudo for all statement after &&

src/ClusterBootstrap/params.py
```diff
@@ -359,7 +359,7 @@
         "CCSAdmins": {
             # The match is in C# Regex Language, please refer to :
             # https://msdn.microsoft.com/en-us/library/az24scfc(v=vs.110).aspx
-            "Allowed": ["jinl@microsoft.com", "hongzl@microsoft.com", "sanjeevm@microsoft.com"],
+            "Allowed": ["jinl@microsoft.com", "hongzl@microsoft.com", "sanjeevm@microsoft.com","zhexu@microsoft.com"],
```
hongzhili: change default allowed to "hongzl, anbhu, jachzh, zhexu, dixu,qixcheng"

src/ClusterBootstrap/params.py
```diff
@@ -387,7 +387,7 @@
     },
 
     "WebUIregisterGroups": ["MicrosoftUsers", "Live", "Gmail"],
-    "WebUIauthorizedGroups": [],  # [ "MicrosoftUsers", "Live", "Gmail" ],
+    "WebUIauthorizedGroups": ["MicrosoftUsers"],  # [ "MicrosoftUsers", "Live", "Gmail" ],
```
hongzhili: keep it as empty.

src/ClusterBootstrap/params.py
```diff
@@ -397,7 +397,7 @@
     "DeployAuthentications": ["Corp", "Live", "Gmail"],
     # You should remove WinBindServers if you will use
     # UserGroups for authentication.
-    "WinbindServers": ["http://onenet40.redmond.corp.microsoft.com/domaininfo/GetUserId?userName={0}"],
+    # "WinbindServers": ["http://onenet40.redmond.corp.microsoft.com/domaininfo/GetUserId?userName={0}"],
```
hongzhili: please check with anbang to make sure we don't need the default winbindservers.

src/ClusterBootstrap/params.py
```diff
@@ -626,27 +631,31 @@
         "runscriptonall ./scripts/prepare_vm_disk.sh",
         "nfs-server create",
         "runscriptonall ./scripts/prepare_ubuntu.sh",
+        "gen_scripts",
```
hongzhili: keep consistent with the changes above.

src/utils/MySQLDataHandler.py
```diff
@@ -14,6 +14,23 @@
 
 from prometheus_client import Histogram
 
+# sys.path.append("../ClusterBootstrap")
```
hongzhili: duplicated codes? consider to move it to the utils.py or somewhere similar so that we can keep it consistent in one place.

src/ClusterBootstrap/deploy.py
YinYangOfDao: gpu-type default value is NULL.
I'm testing it on on-premise devbox; as for on-prem cluster, I probably need the corresponding resource to test.

src/ClusterBootstrap/deploy.py
YinYangOfDao: This is actually not written by me. I think it's Di. I missed this change while editing my own file, so just copied it back.

src/ClusterBootstrap/deploy.py
YinYangOfDao: Di told me he edited the jobmanager, so it's not necessary to run this. I haven't tested yet. If it works, I'd delete this component, otherwise we can discuss next meeting.

src/ClusterBootstrap/install_prerequisites.sh
YinYangOfDao: this is on devbox, one time effort. it's once used to edit /etc/resolv.conf, I'd test whether now we can get rid of it.

src/ClusterBootstrap/deploy.py
YinYangOfDao: do you mean generate these scripts in step gen_configs, using utils.render_template/utils.render_template_directory?

src/ClusterBootstrap/params.py
YinYangOfDao: Confirmed with Anbang, this is internal. Are there any other internal default params that need deleting?

src/ClusterBootstrap/params.py
YinYangOfDao: edited

src/ClusterBootstrap/deploy.py
hongzhili: Yes, please.

src/ClusterBootstrap/deploy.py
hongzhili: Let's delete it.

src/ClusterBootstrap/params.py
Anbang-Hu: Also add "jingzhao,hayua"

src/ClusterBootstrap/az_tools.py
```diff
@@ -600,6 +602,19 @@ def get_disk_from_vm(vmname):
 
     return output.split("/")[-1].strip('\n')
 
+def AZvmsize2GPU(vmsize):
```
Anbang-Hu: Can we make this an Azure mapping file, e.g.
```
#sku gpu-type gpu-count
Standard_NC6s_v2 P100 1
...
```

YinYangOfDao: sure. should I put this mapping file to src/utils, in json format?

YinYangOfDao:
the NFS auto-deployment part is not fully tested yet.

src/ClusterBootstrap/bash_step_by_step_deploy.sh
```diff
@@ -0,0 +1,26 @@
+./deploy.py -y build
```
hongzhili: This can be replaced by script block?

src/ClusterBootstrap/deploy.py
```diff
@@ -58,7 +58,7 @@
 verbose = False
 nocache = False
 limitnodes = None
-
+allroles = {"infra", "infrastructure", "worker", "nfs", "sql"}
```
hongzhili: can we move this var to param.py instead of putting it here as global var?

src/ClusterBootstrap/bash_step_by_step_deploy.sh
YinYangOfDao: sure. but later 2 commands are executed with az_tools.py. do you want to merge them into the script block by running bash cmd in deploy.py? like some "run_command_on_devbox"

PR #488, opened 2019-09-10
Ignore preemptible gpus

hao1939:


src/ClusterManager/dist_pod_template.py
```diff
@@ -101,6 +101,7 @@ def generate_pods(self, job):
         job.data_path = params["dataPath"]
         # TODO user's mountpoints first, but should after 'job_path'
         job.add_mountpoints(job.job_path_mountpoint())
+        job.add_mountpoints({"name": "home", "containerPath": "/home/{}".format(job.get_alias()), "hostPath": job.get_homefolder_hostpath(), "enabled": True})
```
hongzhili: Can you check if "home" mount point is already in the job mountpoint? Only if it's not here, add it.

PR #491, opened 2019-09-11
Expose priority of more statuses of job for new web portal use.

Gerhut:


hao1939:
👍

src/utils/MySQLDataHandler.py
```diff
@@ -998,9 +998,10 @@ def DeleteTemplate(self, name, scope):
             logger.error('Exception: %s', str(e))
             return False
 
+    @record
     def get_job_priority(self):
         cursor = self.conn.cursor()
-        query = "select jobId, priority from {} where jobId in (select jobId from {} where jobStatus in (\"queued\", \"scheduling\", \"running\"))".format(self.jobprioritytablename, self.jobtablename)
+        query = "select jobId, priority from {} where jobId in (select jobId from {} where jobStatus in (\"queued\", \"scheduling\", \"running\", \"unapproved\", \"pausing\", \"paused\"))".format(self.jobprioritytablename, self.jobtablename)
```
hongzhili: This statement may have performance issue if we have large number of jobs in the job table. It should be fine for now. If we hit the performance issue later, this statement can be changed to use joint table selection and build index on foreign key.

PR #498, opened 2019-09-19
support azure blob volume

xudifsd:
For existing k8s cluster, admin should execute `./deploy.py runscriptonall scripts/install-blobfuse.sh` and `./deploy.py kubernetes start flexvolume` to enable this feature. Newly created cluster should already enabled this with normal setup procedure.

Already test in k8s v1.15.2. No user job was impacted.

Later admin should create k8s secret for pod to refer:

```
kubectl create secret generic blobfusecreds --from-literal accountname=ACCOUNT-NAME --from-literal accountkey="ACCOUNT-KEY" --type="azure/blobfuse"
```

And pod looks like:

```
apiVersion: v1
kind: Pod
metadata:
  name: nginx-flex-blobfuse
spec:
  containers:
  - name: nginx-flex-blobfuse
    image: nginx
    volumeMounts:
    - name: test
      mountPath: '/data'
  volumes:
  - name: test
    flexVolume:
      driver: "azure/blobfuse"
      readOnly: false
      secretRef:
        name: blobfusecreds
      options:
        container: 'CONTAINER_NAME'
        mountoptions: "--file-cache-timeout-in-seconds=120"
```

### DIFF
--- a/src/ClusterBootstrap/scripts/install-blobfuse.sh
+++ b/src/ClusterBootstrap/scripts/install-blobfuse.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+sudo rm -f packages-microsoft-prod.deb
+wget https://packages.microsoft.com/config/ubuntu/16.04/packages-microsoft-prod.deb
+sudo dpkg -i packages-microsoft-prod.deb
+sudo apt-get update
+sudo apt-get install -y blobfuse fuse jq
+sudo rm -f packages-microsoft-prod.deb
+sudo mkdir /etc/kubernetes/volumeplugins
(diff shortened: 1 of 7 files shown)


src/ClusterBootstrap/scripts/install-blobfuse.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+sudo rm -f packages-microsoft-prod.deb
+wget https://packages.microsoft.com/config/ubuntu/16.04/packages-microsoft-prod.deb
```
hongzhili: Just want to confirm if 16.04 is correct

xudifsd: Yes, this is correct, they did not provide 18.04 but I have tested it, it works as expected.

PR #500, opened 2019-09-20
Cpuauto

YinYangOfDao:
auto deployment for CPU machine.
image warm up added
modify the create_vm_param rule back so when the role is invalid, it would be counted as "dev" node

### DIFF
--- a/src/ClusterBootstrap/az_tools.py
+++ b/src/ClusterBootstrap/az_tools.py
@@ -523,7 +523,8 @@ def create_vm_param(i, role, vm_size, no_az=False, arm_vm_password=None):
     elif role == "infra":
         vmname = "%s-infra%02d" % (config["azure_cluster"]
                                    ["cluster_name"], i + 1)
-    elif role == "dev":
+    else:
+        role = "dev"
         vmname = "%s-dev" % (config["azure_cluster"]["cluster_name"])
     print "creating VM %s..." % vmname
     vm_ip = get_vm_ip(i, role)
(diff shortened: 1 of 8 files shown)


src/ClusterBootstrap/az_tools.py
```diff
@@ -523,7 +523,8 @@ def create_vm_param(i, role, vm_size, no_az=False, arm_vm_password=None):
     elif role == "infra":
         vmname = "%s-infra%02d" % (config["azure_cluster"]
                                    ["cluster_name"], i + 1)
-    elif role == "dev":
+    else:
+        role = "dev"
```
hongzhili: why is dev role set to default?

src/ClusterBootstrap/params.py
```diff
@@ -655,6 +656,7 @@
         "Standard_NC12s_v2": {"gpu-type": "P100", "gpu-count": 2},
         "Standard_NC24rs_v3": {"gpu-type": "V100", "gpu-count": 4},
         "Standard_NC24rs_v2": {"gpu-type": "P100", "gpu-count": 4},
+        "default": {"gpu-type": "CPU", "gpu-count": 0},
```
hongzhili: Let's call it "None"?, it seems a little bit strange if gpu-type is CPU

YinYangOfDao: sure

src/ClusterBootstrap/az_tools.py
YinYangOfDao: From my perspective, assertion should be a better plan. Anbang pointed out that original code used dev as default, so here using dev as default is consistent with historical code. If you think it fine to use assertion(so we're guaranteed that users select valid machine roles.

PR #503, opened 2019-09-23
Lowpri

YinYangOfDao:
get pribvate IP and render it to  /etc/kubernetes/manifests/kube-apiserver.yaml on infra node.


src/ClusterBootstrap/az_tools.py
```diff
@@ -739,9 +750,10 @@ def gen_cluster_config(output_file_name, output_file=True, no_az=False):
     cc["deploydockerETCD"] = False
     cc["platform-scripts"] = "ubuntu"
     cc["basic_auth"] = "%s,admin,1000" % uuid.uuid4().hex[:16]
+    domain_mapping = {"regular":"%s.cloudapp.azure.com" % config["azure_cluster"]["azure_location"], "low":"redmond.corp.microsoft.com"}
```
hongzhili: domain for low priority cluster should be configurable as well.

src/ClusterBootstrap/params.py
```diff
@@ -655,6 +657,7 @@
         "Standard_NC12s_v2": {"gpu-type": "P100", "gpu-count": 2},
         "Standard_NC24rs_v3": {"gpu-type": "V100", "gpu-count": 4},
         "Standard_NC24rs_v2": {"gpu-type": "P100", "gpu-count": 4},
+        "default": {"gpu-type": "CPU", "gpu-count": 0},
```
hongzhili: change it to None as I commented in another PR.

PR #508, opened 2019-09-24
Add parallel deployment command + Fix IB on Azure installation script

Anbang-Hu:
1. Cast vm names to lower case
2. Reduce worker data disk size from 2047gb to 1023gb
3. Allow overriding `resource_group_name` in config
4. Allow passing `--availability-set` for worker VM creation in Azure
5. Disable unattended package upgrades completely in `disable_kernel_auto_updates.sh`
6. Add a few more ports in default `tcp_port_ranges`
7. Modify IB installation script - `install_ib_on_sriov_az_cluster.sh`
8. `./az_tool.py --parallelism <degree_of_parallelism> create` to create worker VMs in parallel in Azure
9. `./deploy.py runscriptonallinparallel` to run a script all all nodes in parallel
10. `./deploy.py -y updateworkerinparallel` to start kubelet on all worker nodes in parallel

Note: parallel execution is an add-on feature that is completely optional during deployment.
TODO for parallel execution:
1. Aggregate status reporting
2. Error/crash handling
3. Code refactorization

### DIFF
--- a/src/ClusterBootstrap/az_tools.py
+++ b/src/ClusterBootstrap/az_tools.py
@@ -68,8 +68,10 @@ def merge_config(config1, config2, verbose):
 
 
 def update_config(config, genSSH=True):
-    config["azure_cluster"]["resource_group_name"] = config[
-        "azure_cluster"]["cluster_name"] + "ResGrp"
+    if "resource_group_name" not in config["azure_cluster"]:
+        config["azure_cluster"]["resource_group_name"] = config[
+            "azure_cluster"]["cluster_name"] + "ResGrp"
+
     config["azure_cluster"]["vnet_name"] = config[
         "azure_cluster"]["cluster_name"] + "-VNet"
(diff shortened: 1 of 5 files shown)


src/ClusterBootstrap/az_tools.py
```diff
@@ -129,7 +131,7 @@ def create_vm_pwd(vmname, vm_ip, vm_size, use_private_ip, pwd):
                     --nsg %s \
                     --admin-username %s \
                     --storage-sku %s \
-                    --data-disk-sizes-gb 2047 \
+                    --data-disk-sizes-gb 1023 \
```
hongzhili: This is going to override Zhe's PR: https://github.com/microsoft/DLWorkspace/pull/503.
Please revise your change.

Anbang-Hu: Revert this one. To be fixed in another PR.

PR #549, opened 2019-10-09
Expose job status information + do not overwrite existing logs with empty log

Anbang-Hu:


src/ClusterManager/job_manager.py
```diff
@@ -222,6 +304,8 @@ def ApproveJob(job,dataHandlerOri = None):
 
         if "preemptionAllowed" in jobParams and jobParams["preemptionAllowed"] is True:
             logging.info("Job {} preemptible, approve!".format(job_id))
+            detail = [{"message": "waiting for available preemptible resource."}]
```
hongzhili: JobHandler needs to be closed for better performance. I'll add it in my PR later.

PR #551, opened 2019-10-10
use deployment type for inference

hongzhili:


hongzhili:
there is a bug when joblog_manager gets logs from deployment set. we should fix it before next release. @xudifsd

xudifsd:
> there is a bug when joblog_manager gets logs from deployment set. we should fix it before next release. @xudifsd

What bug?

PR #552, opened 2019-10-10
use readiness probe result to accelerate job role status check

xudifsd:
Previous job manager will exec `ls /pod/running/ROLE_READY` in every pod to check if they are ready, this is rather time consuming which will slow loop in job manager.

This PR leverage k8s readiness probe, so that kubelet will execute the command and modify pod readiness, so that job manager can get their readiness in one RPC.

This PR is backward compatible, it will first check if the pod have readiness probe defined, if not, it will fallback to the old way.

### DIFF
--- a/src/ClusterManager/job_manager.py
+++ b/src/ClusterManager/job_manager.py
@@ -366,7 +366,6 @@ def check_job_status(job_id):
     details = []
     for job_role in job_roles:
         details.append(job_role.pod_details().to_dict())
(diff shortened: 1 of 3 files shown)


xudifsd:
Seems working.

image

As more and more job uses readiness, the latency of UpdateJobStatus will decrease.

xudifsd:
We can later wait almost all jobs defined readiness and then add a switch to check the time saved.

PR #557, opened 2019-10-14
move launcher related function into job_launcher

xudifsd:
Moved code in job_deployer/job_role into job_launcher, also moved functions related to launcher from job_manager to job_launcher.

This PR contains only code move. Have tested in dev cluster.


xudifsd:
> A very good start of using framework launch. One suggestion, should we move all DLTS related logic out of PythonLaunch? e.g. job status update, endpoints etc. PythonLaunch may have the similar interface with framework launcher later.

Yes, I want PRs to be as small as possible, and every PR is workable. Will touch that in later PRs.

PR #560, opened 2019-10-15
dashboard: Upgrade use-http to 0.1.91 ...

### LINKED ISSUE #553: useFetch has been updated with breaking changes
I was just taking a look at some of the people using `use-http` and saw this repo. Great work! I just wanted to let you know I deprecated `useGet`, `usePost`, `usePatch`, etc. since all that functionality is baked into `useFetch`.

So, instead of
```jsx
  const { data, error, get } = useGet<Job>({
    url: `/api/clusters/${clusterId}/jobs/${jobId}`,
    onMount: true
  });
```
just do this instead
```jsx
  const { data, error, get } = useFetch<Job>({
    url: `/api/clusters/${clusterId}/jobs/${jobId}`,
    onMount: true
  });
```
Looks like you're only using `useGet` in [4 places](https://github.com/microsoft/DLWorkspace/search?q=use-http&unscoped_q=use-http) so should be a quick fix.

Gerhut:
and remove deprecated methods.

Closes #553 

Kindly ping @LeoHongyi for a code review, tested locally.

P.S. @alex-cory I found that each time the returned `{ get }`-liked methods are different and hard to be listed as dependencies (or it will cause `useEffect` runs in every render), is it by design or my incorrect use?

### DIFF
--- a/src/dashboard/package.json
+++ b/src/dashboard/package.json
@@ -35,7 +35,7 @@
     "recharts": "^1.6.2",
     "typeface-roboto": "^0.0.54",
     "typeface-roboto-mono": "^0.0.54",
-    "use-http": "^0.1.79",
+    "use-http": "^0.1.91",
     "uuid": "^3.3.2"
   },
   "devDependencies": {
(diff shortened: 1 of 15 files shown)


alex-cory:
@Gerhut That sounds like it might be a bug. 🤔I will look into this. I have an idea of what might be causing this. I'll add an issue and get it fixed asap. Thanks for reporting this! 😊

PR #564, opened 2019-10-16
Fix issue of edge incompatibility, adjust the column in job table list, add info to scheduling, and covert to local time in job list job a

LeoHongyi:


yarn.lock
```diff
@@ -0,0 +1,34 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
+"@types/prop-types@*":
```
Gerhut: Could you please merge this file into the one in `src/dashboard` directory?

PR #596, opened 2019-10-25
fix gpu inconsistency between watchdog & restfulapi

xudifsd:
Fix inconsistency between watchdog & restfulapi.


xudifsd:
Some clusters still have inconsistency, some cases:
* one job in West Europe is submitted in August, so it has no vc info attached, watchdog ignore its usage while node manager does not
* restfulapi does not expose used gpu in unschedulable node as unschedulable gpu, but input for calculating quota & available does not get affected, so I do not think this inconsistency is a big deal

PR #617, opened 2019-11-01
add user experience related metrics

xudifsd:
preliminary result in dev1 cluster

image

Left one is result from submitting 200 cpu jobs 1 by 1, right one is result from submitting 50 cpu jobs 1 by 1.

Query: `histogram_quantile(0.95, sum(rate(job_state_change_latency_seconds_bucket[5m])) by (le, current_state))`

`{current_state="submit"}` means 95th latency for job change state from approved to submit to k8s
`{current_state="approve"}` means 95th latency for job state changed from created in restfulapi to approved.

### DIFF
--- a/src/ClusterManager/job_manager.py
+++ b/src/ClusterManager/job_manager.py
@@ -41,6 +41,90 @@ jobmanager_fn_histogram = Histogram("jobmanager_fn_latency_seconds",
             float("inf")),
         labelnames=("fn_name",))
 
+job_state_change_histogram = Histogram("job_state_change_latency_seconds",
+        "latency for job to change state(seconds)",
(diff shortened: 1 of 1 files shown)


src/ClusterManager/job_manager.py
```diff
@@ -41,6 +41,90 @@
...
+        labelnames=("current_state",))
+
+class JobTimeRecord(object):
+    def __init__(self):
+        self.create_time = None
+        self.approve_time = None
+        self.submit_time = None
+
+
+class LRUDefatulDict(object):
```
Anbang-Hu: We might want to leverage third party caching package e.g. https://github.com/tkem/cachetools/tree/c530924cdec86855be6322d3e4dd979bfc9250e4

PR #625, opened 2019-11-03
Add support for ./deploy.py connect samba

Anbang-Hu:


hongzhili:
Do we need to add "samba" in Ln61?

Anbang-Hu:
> Do we need to add "samba" in Ln61?

Added.

PR #633, opened 2019-11-04
add job time from scheduling to running metric

xudifsd:
Add a running state. Also changed name to align with portal, so later we may show these metrics to users.


xudifsd:
related to #617

PR #638, opened 2019-11-05
Inference job: add soft podAffinity to deployment

Gerhut:
Current test result:

Using hard affinity, all pods will be scheduled to one node, until not enough resources.

Using soft affinity, the first 2 pods will be scheduled to the node with the master pod, the 3rd and latter pods will be scheduled to other nodes until all nodes have 2 pods.

### DIFF
--- a/src/Jobs_Templete/deployment.yaml.template
+++ b/src/Jobs_Templete/deployment.yaml.template
@@ -22,11 +22,25 @@ spec:
   replicas: {{ job["deployment_replicas"] }}
   selector:
     matchLabels:
+      jobId: {{ job["jobId"] }}
       worker: active
   template:
     metadata:
       labels:
+        jobId: {{ job["jobId"] }}
         worker: active
+    affinity:
+      podAffinity:
+        preferredDuringSchedulingIgnoredDuringExecution:
+        - weight: 100
+          podAffinityTerm:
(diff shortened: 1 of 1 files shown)


Anbang-Hu:
Even if a cluster is empty (there are a lot of full nodes to occupy), this pod affinity rule will assign only two pods to each node? We would ideally want fragmentation as little as possible.

Gerhut:
I have dug into the scheduler logs and what I found is that the `InterPodAffinityPriority` score is countered by the `SelectorSpreadPriority` score, which are both 10.

Assume there are 2 nodes, one has already one pod scheduled but the other has no.

|                              	| node w/ existing pod 	| node w/o existing pod 	|
|------------------------------	|----------------------	|-----------------------	|
| `BalancedResourceAllocation` 	|                    8 	|                     9 	|
| `LeastResourceAllocation`    	|                    8 	|                     9 	|
| `InterPodAffinityPriority`   	|                   10 	|                     0 	|
| `TaintTolerationPriority`    	|                   10 	|                    10 	|
| `NodeAffinityPriority`       	|                    0 	|                     0 	|
| `SelectorSpreadPriority`     	|                    0 	|                    10 	|
| Sum                          	|                   36 	|                    38 	|

PR #642, opened 2019-11-06
accelerate jobmanager loop

xudifsd:
Submit 50 CPU jobs one by one.

Old job manager:
95th jobs change state
	• from created to approved: 243s
	• from approved to scheduling: 157s
	• from scheduling to running: 124s

Optimized job manager:
95th jobs change state
	• from created to approved: 31s
	• from approved to scheduling: 15s
	• from scheduling to running: 249s


xudifsd:
retired #609

PR #644, opened 2019-11-06
repair manager initial code check-in

debbie-alaine:


src/RepairManager/Rules/ecc_rule.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def get_ECC_error_data(ecc_url):
+
+    response = requests.get(ecc_url)        
+    data = json.loads(response.text)
+    
+    if data:
+        ecc_metrics = data['data']['result']
+        logging.info('ECC error metrics from prometheus ' + json.dumps(ecc_metrics, indent=4))
```
Anbang-Hu: What's the reason for inserting indentation in logs?

src/RepairManager/rule-config.yaml
```diff
@@ -0,0 +1,11 @@
+---  
+# [Rule Module Name] : [Rule Class Name]
+rules:
+  Rules.ecc_rule : ECCRule
+  
+# time to sleep between rule execution
+wait_time: 10
+
+# prometheus ecc error query
+ecc_prometheus_url: 'http://localhost:9091/prometheus/api/v1/query?query=nvidiasmi_ecc_error_count%7Btype%3D%22volatile_double%22%7D%3E0'
```
Anbang-Hu: Should `ecc_prometheus_url` be grouped under `ecc_rule`?

src/RepairManager/main.py
```diff
@@ -0,0 +1,58 @@
...
+
+        # execute all rules listed in config
+        for module_name in rules:
+            try:
+                module = sys.modules[module_name]
+                class_name = rules[module_name]
+                rule_class = getattr(module, class_name)
+                rule = rule_class()
+
+                logger.debug('Executing ' + class_name+ ' from module ' + module_name)
```
Anbang-Hu: Add space after `class_name`.

src/RepairManager/main.py
```diff
@@ -0,0 +1,58 @@
...
+                                
+                if (rule.check_status()):
+                    rule.take_action()
+
+                time.sleep(wait_time)
+
+            except Exception as e:
+                logger.error('Error executing ' + class_name + ' from modul e' + module_name)
```
Anbang-Hu: Remove the extra space in `modul e`.

src/RepairManager/Rules/ecc_rule.py
```diff
@@ -0,0 +1,95 @@
...
+            logging.exception('Error checking status for ECCRule')
+            #TODO: send email alert, raise exception?
+        
+    def take_action(self):
+        try:
+            for node in ecc_hostnames:
+                success = util.cordon_node(node)
+
+                if (success != 0):
+                    logging.warning('Unscheduling of nodes not successful')
```
Anbang-Hu: Can we print which node was cordoned unsuccessfully?

src/RepairManager/Rules/ecc_rule.py
```diff
@@ -0,0 +1,95 @@
...
+            with open('rule-config.yaml', 'r') as rule_config:
+                config = yaml.safe_load(rule_config)
+                       
+            address_map = get_node_address_info()
+
+            ecc_url = config['ecc_prometheus_url']
+            ecc_metrics = get_ECC_error_data(ecc_url)
+    
+            global ecc_hostnames
+            ecc_hostnames = []
```
Anbang-Hu: Can this be a field of `ECCRule`? We can init `self.ecc_hostnames = []` in `def __init__(self)`. Try not to use global variables.

src/RepairManager/rule-config.yaml
hongzhili: it should be prometheus_url in config file (e.g. http://localhost:9091/prometheus or even http://localhost:9091, @xudifsd to confirm). the whole url should be assembled in ecc_rule function.

src/RepairManager/main.py
```diff
@@ -0,0 +1,58 @@
...
+import importlib
+
+import Rules
+
+
+with open('logging.yaml', 'r') as log_file:
+    log_config = yaml.safe_load(log_file)
+
+logging.config.dictConfig(log_config)
+logger = logging.getLogger('basic')
```
xudifsd: Please use `logger = logging.getLogger(__name__)`. Also remove following setLevel, level is already set in logging.yaml config.

src/RepairManager/rule-config.yaml
xudifsd: Yes, prometheus will listen on 9091 port

src/RepairManager/Rules/ecc_rule.py
debbie-alaine: There is no need. I was initially using this when debugging and printing to console. Will remove.

PR #646, opened 2019-11-08
azure blobfuse plugin

leigaoms:
only support 1 blobfuse right now, will support array in next PR.

### DIFF
--- a/src/ClusterManager/job.py
+++ b/src/ClusterManager/job.py
@@ -7,6 +7,8 @@ from jinja2 import Environment, FileSystemLoader, Template
 
 import logging
 import logging.config
+import uuid
+import base64
 
 sys.path.append(os.path.join(os.path.dirname(os.path.abspath(__file__)), "../utils"))
 from osUtils import mkdirsAsUser
@@ -33,6 +35,7 @@ class Job:
                  work_path="",
                  data_path="",
                  params=None,
+                 plugins=None,
                  ):
         """
         job_id: an unique string for the job.
@@ -49,6 +52,7 @@ class Job:
         self.work_path = work_path
         self.data_path = data_path
         self.params = params
+        self.plugins = plugins
 
     def add_mountpoints(self, mountpoint):
         '''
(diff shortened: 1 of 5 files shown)


src/ClusterManager/job.py
```diff
@@ -219,6 +234,35 @@ def _get_cluster_config(self, key):
...
+    def get_plugins(self):
+        '''
+        Plugin example:
+            "blobfuse": {
+                "enabled":true,
+                "name":"blobfuse",
+                "accountName":"YWRtaW4=",
+                "accountKey":"MWYyZDFlMmU2N2Rm",
+                "containerName":"blobContainer",
+                "mountPath":"/data",
```
Anbang-Hu: Can we use another path instead of `/data` as an example? `/data` is currently the cluster-wide shared data path.

src/ClusterManager/job.py
```diff
@@ -219,6 +234,35 @@ def _get_cluster_config(self, key):
...
+                "secreds":"bb9cd821-711c-40fd-bb6f-e5dbc1b772a7"
+                }
+        '''
+        if "plugins" not in self.params.keys():
```
Anbang-Hu: There is no need for `.keys()`. Can we return `None` explicitly for clarity?

src/ClusterManager/job.py
```diff
@@ -219,6 +234,35 @@ def _get_cluster_config(self, key):
...
+            return
+        plugins = self.params["plugins"]
+
+        ret = {}
+        if plugins["blobfuse"] is not None:
```
Anbang-Hu: `"blobfuse"` may not exist in `plugins`.

src/ClusterManager/job.py
leigaoms: Do we certainly have "plugins" in new job? When I run UT tests, old schema jobs does not have "plugins", and cannot use "self.params["plugins"] is None" here.

src/ClusterManager/job.py
leigaoms: yeah, modify it

PR #654, opened 2019-11-12
pull shorter logs

xudifsd:
joblog_manager hanging in pulling logs from jobs with a lot of output, and then later discard most of them. This will shorter the time spent.

### DIFF
--- a/src/utils/k8sUtils.py
+++ b/src/utils/k8sUtils.py
@@ -178,7 +178,7 @@ def GetPod(selector):
     return podInfo
 
 
-def GetLog(jobId):
+def GetLog(jobId, tail=None):
     # assume our job only one container per pod.
 
     selector = "run=" + jobId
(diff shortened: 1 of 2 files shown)


src/utils/k8sUtils.py
```diff
@@ -194,7 +194,7 @@ def GetLog(jobId):
                 if "status" in item and "containerStatuses" in item["status"] and "containerID" in item["status"]["containerStatuses"][0]:
                     containerID = item["status"]["containerStatuses"][0]["containerID"].replace("docker://", "")
                     log["containerID"] = containerID
-                    log["containerLog"] = kubectl_exec(" logs " + log["podName"])
+                    log["containerLog"] = kubectl_exec(" logs %s --tail=%s" % (log["podName"], str(tail)))
```
hongzhili: Maybe we should allow to get all logs if tail = None
if tail is not None:
    log["containerLog"] = kubectl_exec(" logs %s --tail=%s" % (log["podName"], str(tail)))
else:
    log["containerLog"] = kubectl_exec(" logs " + log["podName"])

xudifsd: ok, fair enough.

PR #656, opened 2019-11-13
profile more

xudifsd:
profile more calls


hongzhili:
@xudifsd , src/ClusterManager/job_launcher.py is using "logger" and "logging" in the same time, which one should be used?

xudifsd:
> @xudifsd , src/ClusterManager/job_launcher.py is using "logger" and "logging" in the same time, which one should be used?

We should use logger, I still do not have time to change it. I will submit a dedicated PR to change all of the logging.

PR #666, opened 2019-11-19
fix init script 

hongzhili:


hongzhili:
This is to support distributed job, if the docker image has \#Port 22 in the ssh config file. @xudifsd @Anbang-Hu , please have a look.

PR #668, opened 2019-11-20
support password login

hongzhili:


src/ClusterManager/job_launcher.py
```diff
@@ -523,6 +524,17 @@ def submit_job_impl(self, job):
...
 
+            if "job_token" not in job_object.params:
+                if "user_sign_token" in config and "userName" in job_object.params:
+                    job_object.params["job_token"] = hashlib.md5(job_object.params["userName"]+":"+config["user_sign_token"]).hexdigest()
+                else:
+                    job_object.params["job_token"] = "tryme2017"
+
+            if "envs" not in job_object.params:
+                job_object.params["envs"] =[]
+                job_object.params["envs"].append({"name": "DLTS_JOB_TOKEN", "value": job_object.params["job_token"]})
```
xudifsd: I think this line should not be in `if` block?

src/ClusterManager/job_launcher.py
```diff
@@ -523,6 +524,17 @@ def submit_job_impl(self, job):
             job_object.params["uid"] = user_info["uid"]
             job_object.params["user"] = job_object.get_alias()
 
+            if "job_token" not in job_object.params:
+                if "user_sign_token" in config and "userName" in job_object.params:
```
Anbang-Hu: Just want to make sure that this is an email, not just an alias.

src/ClusterManager/job_launcher.py
hongzhili: right, good catch

src/ClusterManager/job_launcher.py
hongzhili: it is email in the current implementation. maybe we need to force the format in portal or restfulapi

PR #675, opened 2019-11-22
support db pool

leigaoms:
provide thread-safe mysql connection pool for two libraries, and add config to switch different pools


leigaoms:
> Can we inherit these two new handler classes from MySQLDataHandler to share common code?



> Can we inherit these two new handler classes from MySQLDataHandler to share common code?

MySQLDataHandler is not threadsafe, it initializes connector in constructor, and closes connection in a separate close(). But in DBPool, I get and close connection in each function, and the pool is a singleton. So inheriting will need many changes on MySQLDataHanlder, and make it hard to rollback. I will clean up the duplicated code after the pool is stabilized on production later.

PR #679, opened 2019-11-24
support ldap auth

hongzhili:


src/ClusterBootstrap/services/jobmanager/pre-render.sh
```diff
@@ -1,7 +1,6 @@
 #!/bin/bash
 
 dir=`dirname $0`
-
-dlws_scripts_file_name=${dir}/dlws-scripts.yaml
-
-${dir}/../../deploy/bin/kubectl create configmap dlws-scripts --from-file=${dir}/../../../init-scripts --dry-run -o yaml > $dlws_scripts_file_name
+dlws_scripts_file_name=./deploy/${dir}/dlws-scripts.yaml
+./deploy.py rendertemplatedirectory ../init-scripts ./deploy/init-scripts
```
Anbang-Hu: Does this line work properly?

hongzhili: it should work. Do you see any issue?

xudifsd: I think we'd better render init-script into one yaml file, and then render the template in yaml file. So in my opinion, we do not need to call rendertemplatedirectory at all here. Previous version should just works.

And BTW, this is pre-render.sh, it happens before `deploy` actual created, so if you pointing `dlws_scripts_file_name` to `deploy`, it will either fail in the first run, or will create configmap from previous rendered files.

hongzhili: rendertemplatedirectory is used to populate some config values to init_script. and they are rendered into a single yaml file by kubectl. 
indeed, there is an issue to call this function before deploy.py build or deploy.py download. Zhe is going to fix it in his PR.

PR #681, opened 2019-11-25
Handle terminating pods when machines are taken away for k8s > 1.13 +  expose more job info

Anbang-Hu:
1. Handle terminating pods when machines are taken away - [Issues 72226](https://github.com/kubernetes/kubernetes/issues/72226)
2. Add check for `config["user_sign_token"]` being `None`
3. Print `node_name`, `host_ip`, `pod_ip` for jobs in restricted detail
4. Expose requested and available GPU resources for user visibility in queued jobs

### DIFF
--- a/src/ClusterManager/job_launcher.py
+++ b/src/ClusterManager/job_launcher.py
@@ -535,7 +535,7 @@ class PythonLauncher(Launcher):
             job_object.params["user"] = job_object.get_alias()
 
             if "job_token" not in job_object.params:
-                if "user_sign_token" in config and "user_sign_token" is not None and "userName" in job_object.params:
+                if "user_sign_token" in config and config["user_sign_token"] is not None and "userName" in job_object.params:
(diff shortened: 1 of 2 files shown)


src/ClusterManager/job_launcher.py
```diff
@@ -369,17 +369,34 @@ def status(self, refresh=False):
...
+                return "Unknown"
+
             # Check if the user command had been ran.
             if not self._is_role_ready():
                 return "Pending"
 
         return phase
 
+    def pod_restricted_details(self):
+        detail = dict()
```
xudifsd: Use literal value over constructor function: https://doughellmann.com/blog/2012/11/12/the-performance-impact-of-using-dict-instead-of-in-cpython-2-7-2/

Anbang-Hu: Nice.

PR #687, opened 2019-11-26
User synchronizer

Gerhut:
TODO:

- [x] deploy

### DIFF
--- a/src/ClusterBootstrap/services/user-synchronizer/cronjob.yaml
+++ b/src/ClusterBootstrap/services/user-synchronizer/cronjob.yaml
@@ -1,0 +1,35 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: user-synchronizer-cronjob
+spec:
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    metadata:
+      name: user-synchronizer-job
+    spec:
+      template:
+        metadata:
+          name: user-synchronizer-pod
+        spec:
+          containers:
+          - name: user-synchronizer
+            image: '{{cnf["worker-dockerregistry"]}}{{cnf["dockerprefix"]}}user-synchronizer:{{cnf["dockertag"]}}'
+            imagePullPolicy: Always
+            env:
+            - name: TENANT_ID
+              value: '{{cnf["user-synchronizer"]["tenant_id"]}}'
+            - name: CLIENT_ID
+              value: '{{cnf["user-synchronizer"]["client_id"]}}'
+            - name: CLIENT_SECRET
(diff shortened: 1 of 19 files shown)


src/ClusterBootstrap/services/user-synchronizer/cronjob.yaml
```diff
@@ -0,0 +1,35 @@
...
+              value: '{{cnf["user-synchronizer"]["client_secret"]}}'
+            - name: WINBIND_URL
+              value: '{{cnf["user-synchronizer"]["winbind_url"]}}'
+            - name: DATABASE_URL
+              value: 'mysql://{{cnf["mysql_username"]}}:{{cnf["mysql_password"]}}@{{cnf["mysql_node"]}}:{{cnf["mysql_port"]}}/DLWSCluster-{{cnf["clusterId"]}}'
+            - name: GROUPS_ID
+              value: '{{cnf["user-synchronizer"]["groups_id"]}}'
+          nodeSelector:
+            user-synchronizer: active
+          restartPolicy: Never
```
Anbang-Hu: Do we not want to restart upon failure?

src/user-synchronizer/logger.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+from logging import getLogger, FileHandler, StreamHandler
```
Anbang-Hu: Let's use configuration file: https://github.com/microsoft/DLWorkspace/blob/dltsdev/src/RestAPI/logging.yaml

src/user-synchronizer/README.md
```diff
@@ -0,0 +1,23 @@
...
+## Get Started
+
+```shell
+pipenv install
+pipenv run start
+```
+
+## Configuration
+
+It uses environment variables to configure the behavior of the service:
```
Anbang-Hu: Not sure which is a better option - using environment variables or using a yaml configuration file.

src/user-synchronizer/__main__.py
```diff
@@ -0,0 +1,76 @@
...
+        'gid': int(identity['gid']),
+        'groups': dumps(identity['groups'], separators=(',', ':')),
+    }
+    response = get(restfulapi_url + '/AddUser', params=params)
+    response.raise_for_status()
+
+    logger.info('Updated {} to cluster'.format(user_name))
+
+
+for group_identity in iter_group_identities():
```
Anbang-Hu: Periodic run?

src/user-synchronizer/README.md
Gerhut: Either is OK, I am just following [the 12 factors](https://12factor.net/config)

PR #695, opened 2019-12-03
sync and service discovery using k8s

xudifsd:
Previously, we use NFS to do sync and service discovery, this is the first step to remove dependency on NFS during init.

This PR doesn't removed NFS dependency, logs and hostfiles are still generated on NFS.


Anbang-Hu:
Can we have consistent `/etc/ssh/sshd_config` with `indexserveregistry.azurecr.io/deepscale:2.0`? Especially allowing password SSH, X11 Forwarding and `ClientAliveInterval 30`.

src/ClusterManager/job_launcher.py
```diff
@@ -65,6 +65,36 @@ def _delete_pod(self, name, grace_period_seconds=None):
...
+            message = "Delete pods failed: {}".format(label_selector)
+            logging.warning(message, exc_info=True)
+            errors.append({"message": message, "exception": e})
+        return errors
+
+    @record
+    def _cleanup_configmap(self, label_selector):
+        errors = []
+        try:
+            api_response = self.k8s_CoreAPI.delete_collection_namespaced_config_map(
```
Anbang-Hu: `api_response` seems unused

xudifsd:
> ClientAliveInterval

addressed in #700 , I will rebase to dltsdev after 700 merged, so that we can rollback this feature quickly

PR #697, opened 2019-12-03
repair manager: more refactoring

debbie-alaine:
cleaned up email handling, organized config and util files. also added mock data for potential testing.

updated prebuild.sh so that it can render a config with email credentials before building docker container.


src/RepairManager/Rules/ecc_rule.py
```diff
@@ -45,6 +39,10 @@ def get_ECC_error_data(ecc_url):
 
     response = requests.get(ecc_url)        
     data = json.loads(response.text)
+
+    # use mock data
```
Anbang-Hu: Remove commented code if no longer needed.

src/RepairManager/config/rule-config.yaml
```diff
@@ -0,0 +1,9 @@
+---  
+rules:
+        ecc_rule:
```
Anbang-Hu: Can we use two-space indentation in yaml files?

src/RepairManager/Rules/ecc_rule.py
```diff
@@ -101,8 +100,6 @@ def take_action(self):
                 else:
                     body += f'{node_name}: Successfully marked as unschedulable\n'
 
-        if not all_nodes_already_unscheduled:
```
Anbang-Hu: Looks like `all_nodes_already_unscheduled` is no longer used. Consider deleting?

src/RepairManager/utils/email.py
```diff
@@ -0,0 +1,52 @@
...
+    def __init__(self):
+        self.config=self.load_config()
+        self.monitor_alerts = {}
+
+    def load_config(self):
+        with open('./config/email-config.yaml', 'r') as file:
+            return yaml.safe_load(file)
+
+    def send(self, subject, body):
+        message = "From: %s\r\nTo: %s\r\nSubject: %s\r\n\r\n%s" % (self.config['sender'], self.config['receiver'], subject, body)
```
Anbang-Hu: Can we use `\n` instead of `\r\n`?

PR #700, opened 2019-12-04
use deepscale sshd config

xudifsd:
use deepscale sshd config


hongzhili:
@xudifsd, I assume we should keep the changes in #700 ?

PR #702, opened 2019-12-05
repair manager email fixes

debbie-alaine:
some fixes related to email configuration/handling in repair manager


src/RepairManager/Rules/ecc_rule.py
```diff
@@ -88,10 +88,16 @@ def take_action(self):
```
Anbang-Hu: `if not k8s_util.is_node_unschedulable(self.node_info, node_name):` means if node is schedulable, cordon node?

PR #706, opened 2019-12-06
Use batch delete secrets

Anbang-Hu:


src/dashboard/src/pages/Submission/Training.tsx
```diff
@@ -1021,7 +1021,7 @@ const Training: React.ComponentClass = withRouter(({ history }) => {
                 </Grid>
                 <Grid item xs={12}>
                   <TextField
-                    value={mountOptions}
+                    defaultValue={mountOptions}
```
Gerhut: What's this for?

xudifsd:
We can also use this way to delete pods/services and deploymenets

Anbang-Hu:
> We can also use this way to delete pods/services and deploymenets

I am only handling my code now. pod/service/deployment cleanup refactor for performance improvement should come in a different PR.

PR #713, opened 2019-12-13
Fix dashboard deployment

Gerhut:


hongzhili:
May I get the context of this change? Is winbindserver a list now?

Gerhut:
> May I get the context of this change? Is winbindserver a list now?

Yes, `winbind` field in config.yaml is always list, which should be a single string here

Anbang-Hu:
Can we parameterize these two configs as well?
```
AddGroupLink
```

Gerhut:
> Can we parameterize these two configs as well?
> 
> ```
> AddGroupLink
> WikiLink
> ```

It's restructured in `https://github.com/microsoft/DLWorkspace/pull/714`, will open in a new PR

PR #718, opened 2019-12-20
Dashboard backend: adjust some logs to debug level

Gerhut:


src/dashboard/server/api/configurations/logger.js
```diff
@@ -5,10 +5,14 @@ const logger = require('koa-pino-logger')
...
+  const level = {
+    test: 'silent',
+    development: 'debug',
+    production: 'info'
+  }[app.env]
   const prettyPrint = app.env === 'development'
-  const middleware = logger({ enabled, prettyPrint })
-  app.silent = !enabled
+  const middleware = logger({ level, prettyPrint })
+  app.silent = app.env !== 'test'
```
Anbang-Hu: Just want to make sure this is the logic we want, because it looks different from before.
```
app.silent = app.env !== 'test'
```

It was
```
app.silent = !(app.env !== 'test')
```

Gerhut: Sorry, my bad. And `app.silent` is not covered in our unit tests

PR #724, opened 2019-12-29
restrict port range for ssh

xudifsd:
restrict to 40000-49999 inclusively.

### DIFF
--- a/src/docker-images/init-container/runtime/sync.py
+++ b/src/docker-images/init-container/runtime/sync.py
@@ -8,6 +8,7 @@ import socket
 from contextlib import closing
 import json
 import time
+import random
 from pathlib import Path
 
 from kubernetes import client as k8s_client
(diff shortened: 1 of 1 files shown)


src/docker-images/init-container/runtime/sync.py
```diff
@@ -22,13 +23,24 @@
...
         "k8s_api": 3,
+        "port": 4,
         }
 
-def find_free_port():
-    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
-        s.bind(("", 0))
-        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-        return s.getsockname()[1]
+def find_free_port(min=30000, max=49999):
```
Anbang-Hu: Let's restrict it to 40000-49999 to be backward compatible. https://github.com/microsoft/DLWorkspace/blob/617d14f1b8634ff6cbe576f3bf8f3cac52047c89/src/init-scripts/setup_sshd.sh#L32

xudifsd: done

PR #735, opened 2020-01-04
repairmanager - do not resend email when all nodes are already cordoned

debbie-alaine:
When turning off dry-run, two emails are sent. This is because after nodes are cordoned, their status changes to "already cordoned." Since this is a status change, an additional email is sent.

This update
- checks if nodes are already cordoned before attempting to cordon
- does not send emails if all nodes listed have already been cordoned in the past.

Note: this update interferes with the "reminder" functionality... will make a fix in future PR.


debbie-alaine:
Still a few bugs in email alerting I need to work through. Closing and will send a new one.

PR #743, opened 2020-01-08
Send to CC list and refactor storagemanager

Anbang-Hu:


src/StorageManager/path_tree.py
```diff
@@ -52,7 +52,8 @@ def _walk(self, root):
         try:
             pathnames = os.listdir(root)
         except Exception as e:
-            self.logger.warning("Ignore path %s due to exception %s" % (root, e))
+            self.logger.warning("Ignore path %s due to exception %s" %
```
xudifsd: please use `self.logger.warning("xxx %s %d", yyy, zzz)`, do not format the logging string first before calling warning, one of reason people use logger is when logger is disabled, this style will have less performance impact than yours.

Same below.

src/StorageManager/storage_manager.py
```diff
@@ -55,62 +63,100 @@ def run(self):
...
+        """Returns the used percentage number for the specified scan_point.
+
+        Args:
+            scan_point: Configuration for a mountpoint to scan.
+
+        Returns:
+            Used percentage number of the scan_point or None otherwise.
+        """
+        try:
+            df = subprocess.Popen(["df", scan_point["device_mount"]],
```
xudifsd: You may want to add a timeout to Popen.

PR #717, opened 2019-12-20
perf optimization for job list and detail

leigaoms:
1. Add ListJobsV2 API: merge 2 DB requests into 1 by simplifying code logic; cut useless fields; add job priority when getting all jobs.
2. Add GetJobDetailV2 API: cut useless fields; separate endpoints acquisition.
3. Add UpdateJobTextFields in DataHandler: job manager could update multiple fields of a job in one request
4. Optimize authorization cache: use fine-grained key (eg. identityName, resource), so that cache could be updated when DB updates, and do not have to wait until TTL expired
5. Optimize VC list cache: use vcName as key, instead of userName. Update cache timely after DB is successfully updated
6. Improve Enpoint API: Move DB related logic into JobRestAPIUtils, to prevent restAPI from depending on DataHandler directly. Simplify DB operation to reduce latency

PR #746, opened 2020-01-09
fix endpoint extract

xudifsd:
related to #717


xudifsd:
Should cherry-pick to r1.4

PR #747, opened 2020-01-09
repair manager - email alerting refactoring

debbie-alaine:
Migrated some of the alerting logic to RuleAlertHandler to handle caching and reminders, attempting to decouple from any specific rule.

The RuleAlertHandler now includes a ttl-cache window, so that if a node has not been seen for a configured time (currently set to one hour), it will send an alert email. this was an issue in one of the clusters were the prometheus query was not consistent in the nodes (switching from 2 nodes with ecc error, to 3, then back to 2 every couple of minutes). This was causing many emails.


debbie-alaine:
@Anbang-Hu  @hongzhili  @YinYangOfDao let me know if you have any feedback.

src/dev-utils/prometheus_url.py
```diff
@@ -1,47 +1,47 @@
...
-
-    parser.add_argument("--since", "-s", type=int, default=ago,
-            help="start time for generating report")
-    parser.add_argument("--until", "-u", type=int, default=now,
-            help="end time for generating report")
-
-    args = parser.parse_args()
-
-    main(args)
+#!/usr/bin/env python3
```
Anbang-Hu: Why is this file showing diff?

src/RepairManager/test/test_ecc_rule.py
```diff
@@ -0,0 +1,253 @@
+import os, sys
```
Anbang-Hu: Can you follow #769 to enable tests for repairmanager?

src/dev-utils/prometheus_url.py
debbie-alaine: oops, should not have updated here. though I did update it in common utils in Repair Manager. will fix.

PR #750, opened 2020-01-09
change cluster_manager, restfulapi and deploy.py to python3

xudifsd:
Most of changes are made by `2to3` program.

Also removed a lot of dead code in deploy.py.

Have tested cluster manager and restful api. Works, but not sure if we will meet some problem.


src/ClusterBootstrap/gs_tools.py
```diff
@@ -347,13 +347,13 @@ def create_storage_with_config( configGrp, location ):
     storagename = configGrp["name"] + location
     output = create_storage_account( storagename, configGrp["sku"], location)
     if verbose: 
-        print ( "Storage account %s" % output )
+        print(( "Storage account %s" % output ))
```
Anbang-Hu: double brackets

xudifsd: Yes, I see many of this. I think it's because 2to3 will add brace whenever it encountered a `print`

xudifsd: Will delete these

src/ClusterBootstrap/az_tools.py
```diff
@@ -53,15 +53,15 @@ def merge_config(config1, config2, verbose):
         if entry in config1:
             if isinstance(config1[entry], dict):
                 if verbose:
-                    print("Merge entry %s " % entry)
+                    print(("Merge entry %s " % entry))
```
Anbang-Hu: double braces

xudifsd:
after this, we can use https://pypi.org/project/autopep8/ to format all python code

PR #752, opened 2020-01-09
User synchronizer: filter out subgroups

Gerhut:


Anbang-Hu:
@Gerhut please cherry-pick to r1.4

PR #772, opened 2020-01-17
Define new resource type to simplify code logic

Anbang-Hu:
TODO
- [x] Add test cases for `ClusterStatus`
- [x] Enable tests in `.travis.yml`


src/ClusterManager/node_manager.py
```diff
@@ -73,7 +79,7 @@ def get_job_gpu_usage(jobId):
     return gpuUsage
 
 
-def get_cluster_status():
+def _get_cluster_status():
```
hongzhili: Let's remove the codes eventually.

PR #781, opened 2020-01-22
Expose CPU and memory usage in ClusterStatus

Anbang-Hu:


src/ClusterManager/cluster_status.py
```diff
@@ -214,46 +272,77 @@ def __parse_pod_statuses(self):
                 if gpu_usage <= 25:
                     pod_name += "!!!!!!"
```
hongzhili: Let's change this hacking code later

PR #782, opened 2020-01-22
Remove Legacy Codes

hongzhili:


hongzhili:
Please be careful on this PR. A lot of deletions may break the current workflow

PR #811, opened 2020-02-08
Auto expiration of paths in storage manager and refactor

Anbang-Hu:
Different storage manager rules are added in this PR. Each `scan_point` produces a `tree`, and then process these rules (if they are enabled)
- `OverweightRule`:  (__enabled__ by default) Find over-sized boundary paths and send email alerts. Only triggers when actual usage `>= used_percent_alert_threshold`. `used_percent_alert_threshold` can be configured per `scan_point` and its default is `200G`. It can be turned off by setting `overweight_rule: False` per `scan_point`.
- `ExpiredRule`:  (__disabled__ by default) Find expired boundary paths and send email alerts. A path is expired when its last access time is older than `expiry_days`. The default value of `expiry_days` is `31`. It can be turned on by setting `expired_rule: True` for each `scan_point`.
- `ExpiredToDeleteRule`: (__disabled__ by default) Delete expired boundary paths. An expired path is deleted when its last access time is older than `expiry_days + days_to_delete_after_expiry`. The default value of `days_to_delete_after_expiry` is `None`. It can be set per `scan_point`. It can be turned on by setting `expired_to_delete_rule: True` per `scan_point`.
- `EmptyRule`: (__disabled__ by default) Find empty boundary directories and send email alerts. It can be turned on by setting `empty_rule: True` per `scan_point`.

For a `ExpiredToDeleteRule` to execute, both conditions have to be met:
1. `expired_to_delete_rule` is `True`
2. `days_to_delete_after_expiry` should not be `None`

Say I want to start detecting paths older than 7 days and auto delete paths older than 14 days under `/data/share/storage/experiments`. In `config.yaml`:
```
storagemanager:
  nfs_storage1:

    #Default values
    #expiry_days: 31
    #days_to_delete_after_expiry: None
    #used_percent_alert_threshold: 214748364800

    scan_points:
    - path: /data/share/storage/experiments
      alias: /data/experiments

      #Default overweight_rule value
      #overweight_rule: True

      expiry_days: 7
      expired_rule: True

      days_to_delete_after_expiry: 7
      expired_to_delete_rule: True

      #Default empty_rule value
      #empty_rule: True
```

### DIFF
--- a/src/StorageManager/rule.py
+++ b/src/StorageManager/rule.py
@@ -42,7 +42,7 @@ class Rule(object):
                            self.path)
             return
 
-        logger.info("Rule % is enabled for %s. Processing ...",
+        logger.info("Rule %s is enabled for %s. Processing ...",
                     self.name,
                     self.path)
         self.run_rule()
@@ -137,15 +137,17 @@ class Rule(object):
 
     def delete_nodes(self):
         if len(self.nodes) == 0:
-            logger.info("No %s nodes to delete.", )
+            logger.info("No nodes to delete for rule %s.", self.name)
             return
 
-        logger.info("Deleting nodes ...")
+        logger.info("Deleting nodes for rule %s...", self.name)
         for node in self.nodes:
             if os.path.exists(node.path):
                 if node.isdir:
+                    logger.info("Deleting directory %s...", node.path)
                     os.rmdir(node.path)
                 else:
+                    logger.info("Deleting file %s...", node.path)
                     os.remove(node.path)
             else:
                 logger.warning("%s does not exist", node.path)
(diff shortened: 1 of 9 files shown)


src/StorageManager/rule.py
```diff
@@ -0,0 +1,294 @@
...
+        for node in self.nodes:
+            if os.path.exists(node.path):
+                if node.isdir:
+                    logger.info("Deleting directory %s...", node.path)
+                    os.rmdir(node.path)
+                else:
+                    logger.info("Deleting file %s...", node.path)
+                    os.remove(node.path)
+            else:
+                logger.warning("%s does not exist", node.path)
```
hongzhili: Maybe we should add try/catch in side for loop. In case of, although it may be rare, file is deleted between os.path.exists and os.rmdir

Also, is it possible to reduce the risk of doing os.rmdir ("/*") or something similar.....

PR #812, opened 2020-02-08
add test cases for azure blob

xudifsd:


src/ClusterManager/test/test_distributed_job.py
```diff
@@ -142,3 +144,45 @@ def test_distributed_job_env(args):
...
+                                           ps_container_name, ps_cmd)
+        assert code == 0, "code is %d, output is %s" % (code, output)
+
+        worker_label = "jobId=%s,jobRole=worker" % job.jid
+        pods = utils.kube_get_pods(args.config, "default", worker_label)
+        assert len(pods) == 1
+
+        worker_pod_name = pods[0].metadata.name
+        worker_container_name = pods[0].spec.containers[0].name
+        worker_cmd = ["bash", "-c", "cat /tmp/blob/${DLWS_JOB_ID}"]
```
Anbang-Hu: Shall we remove the file in worker command?

src/ClusterManager/test/utils.py
```diff
@@ -159,16 +102,31 @@ def post_inference_job(rest_url, email, uid, vc, image, cmd):
         "jobPath": "",
         "enablejobpath": True,
         "env": [],
-        "hostNetwork": False,
-        "isPrivileged": False,
-        "resourcegpu": 1,  # num of worker
+        "resourcegpu": 0,
+        "cpulimit": 1,
```
Anbang-Hu: Why adding `cpulimit`?

PR #784, opened 2020-01-23
Elastic search based job logging (from repo Gerhut/DLWorkspace)

Gerhut:
In `config.yaml`, add
```
azure_cluster:
  elasticsearch_node_num: 3
  #elasticsearch_vm_size: Standard_DS2_v2
  #elasticsearch_local_storage_sz: 2048
  #vm_image: Canonical:UbuntuServer:18.04-LTS:18.04.201912180
  #azure_location: eastus
```

Deployment commands:
```
./az_tools.py create
./az_tools.py genconfig
./deploy.py runscriptonroles elasticsearch ./scripts/prepare_vm_disk.sh
./deploy.py runscriptonroles elasticsearch ./scripts/prepare_ubuntu.sh
./deploy.py runscriptonroles elasticsearch ./scripts/disable_kernel_auto_updates.sh
./deploy.py runscriptonroles elasticsearch ./scripts/docker_network_gc_setup.sh
./deploy.py runscriptonroles elasticsearch ./scripts/disable_mlocate.sh
./deploy.py genscripts
./deploy.py runscriptonroles elasticsearch ./scripts/dns.sh
./deploy.py -y updateelasticsearch
./deploy.py -y kubernetes labels
./deploy.py kubernetes start logging
```

PR #815, opened 2020-02-10
fix decode issue for old pipeline

YinYangOfDao:


Anbang-Hu:
We've fixed this in #784. If you think the fix is inappropriate, can you resolve the conflict and push again?

YinYangOfDao:
Colleagues fixed this in #784, with bytes instance checking.

PR #824, opened 2020-02-12
Fix log test with cursor

Gerhut:
Azure blob & env not tested


xudifsd:
@Gerhut I have tested in dev1, but still get empty log. Not sure why. I have changed the deployment files and added you as reviewer, you can sync with it so you can have test on azure blob.

PR #832, opened 2020-02-14
try to use controller and fixed some inconsistency

xudifsd:
our job manager start pod has different name than framework controller, this requires some changes in endpoint manager.
Still can not use framework controller now, because it doesn't generate unique label so Service generated by endpoint manager can not have a valid endpoint.

Have asked Yuqi to implement this unique label. And then I will change code accordingly.

Apart from this, I also fix our test case, previous distributed job test environment by kubectl exec into pod, but some dynamic generated environment can not be tested. So changed the test case by ssh into the pod.


xudifsd:
need https://github.com/microsoft/frameworkcontroller/pull/53

PR #835, opened 2020-02-17
Fluentd: explicitly set refresh_interval to 1s instead of 60s by default.

Gerhut:
This change will fix log missing if the job is finished in less than 60 seconds.

May affect I/O performance.


src/ClusterBootstrap/services/logging/fluentd-config.yaml
```diff
@@ -0,0 +1,293 @@
...
+    <filter kubernetes.**>
+      @type kubernetes_metadata
+      @id filter_kube_metadata
+    </filter>
+  prometheus.conf: |
+    # Prometheus metric exposed on 0.0.0.0:24231/metrics
+    <source>
+      @type prometheus
+      bind "#{ENV['FLUENTD_PROMETHEUS_BIND'] || '0.0.0.0'}"
+      port "#{ENV['FLUENTD_PROMETHEUS_PORT'] || '24231'}"
```
xudifsd: Is this the same to `{{cnf["fluentd"]["port"]}}`?

Gerhut: It's overwritten in envvar, which is `{{cnf["fluentd"]["port"]}}`

See [#diff-a40d28c47725cc72f15c36d7ae3f90eaR37-R38]()

Anbang-Hu:
@Gerhut please help resolve the conflict and merge

Gerhut:
#836 includes this, so this could be closed.

PR #839, opened 2020-02-18
support inference job in launcher

xudifsd:
All work for introduce frameworker are done. But when switching to new launcher, we can not simply change it, because old running jobs can not use new one to manage.

Here are choices we can make:

1. write some auto detection so new job will use new launcher to manage, old ones will be managed by old launcher
2. use one launcher per cluster, and only enable this new launcher in new cluster, and migrate old cluster when we can

The first will need to write some code, it's not trivial. The second only need to add a switcher, is much easier to implement. @hongzhili @Anbang-Hu what's your opinion?

### DIFF
--- a/src/ClusterManager/framework.py
+++ b/src/ClusterManager/framework.py
@@ -41,7 +41,7 @@ class Framework(object):
                  home_folder_host_path, is_fragment_gpu_job,
                  is_preemption_allowed, is_host_network, is_host_ipc,
                  is_privileged, is_nccl_ib_disabled):
-        self.init_image = init_image  # str
+        self.init_image = init_image  # str, maybe None
         self.labels = labels  # map
         self.annotations = annotations  # map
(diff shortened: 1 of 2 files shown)


src/ClusterManager/framework.py
```diff
@@ -728,5 +735,79 @@ def transform_distributed_job(params, cluster_config):
...
+    )
+    worker_resource = transform_resource(
+        params,
+        cluster_config.get("default_cpurequest", "500m"),
+        cluster_config.get("default_cpulimit", "500m"),
+        cluster_config.get("default_memoryrequest", "2048M"),
+        cluster_config.get("default_memorylimit", "2560M"),
+    )
+    master_resource.gpu_limit = 0
+    worker_resource.gpu_limit = 0
```
Anbang-Hu: 1worker_resource.gpu_limit1 may not be 0

PR #840, opened 2020-02-19
Generalize cluster resource

Anbang-Hu:
1. Resource should always be non-negative
2. Replace `gpuType` with `sku` in resource calculation in `job_manager.TakeJobActions`.
3. Insert default `sku` value if it's not specified upon job submission.

TODO:
1. Expose `sku` instead of `gpuType` in `GetVC`.
2. Save `pods` in `ClusterStatus`.
2. Refactor VC cluster info in `VirtualClusterStatus`.


src/utils/resource_stat.py
```diff
@@ -174,52 +218,61 @@ def __imul__(self, other):
...
+
+            result.normalize()
             return result
 
     def __idiv__(self, other):
         if isinstance(other, numbers.Number):
             for k, v in self.res.items():
-                self.res[k] = v / other
+                # Division by zero gives zero
+                if other == 0:
```
xudifsd: do we need to log here and other div functions? because we are essentially swallowing the exception

PR #846, opened 2020-02-20
add switcher in job manager to switch launcher

xudifsd:
Will use `dlws/frameworkcontroller` as controller image, because upstream do not have [fix](https://github.com/microsoft/frameworkcontroller/pull/53#issuecomment-588008410) we needed. Will change to upstream image once it's released.

### DIFF
--- a/src/ClusterBootstrap/services/launcher/launcher.yaml
+++ b/src/ClusterBootstrap/services/launcher/launcher.yaml
@@ -19,4 +19,4 @@ spec:
       # serviceAccountName: frameworkcontroller
       containers:
       - name: frameworkcontroller
-        image: xudifsd/frameworkcontroller
+        image: dlws/frameworkcontroller
--- a/src/ClusterBootstrap/services/launcher/launcher.yaml
+++ b/src/ClusterBootstrap/services/launcher/launcher.yaml
@@ -19,4 +19,4 @@ spec:
       # serviceAccountName: frameworkcontroller
       containers:
       - name: frameworkcontroller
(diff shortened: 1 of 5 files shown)


src/ClusterBootstrap/services/launcher/launcher.yaml
```diff
@@ -19,4 +19,4 @@ spec:
       # serviceAccountName: frameworkcontroller
       containers:
       - name: frameworkcontroller
-        image: frameworkcontroller/frameworkcontroller:v0.5.1
+        image: xudifsd/frameworkcontroller
```
Anbang-Hu: We probably should not use personal identifier in repo?

src/docker-images/RestfulAPI/runScheduler.sh
```diff
@@ -1,7 +1,5 @@
 #!/bin/bash
-rm /DLWorkspace/src/utils/config.yaml
```
Anbang-Hu: Does `ln` replace the old target?

src/ClusterBootstrap/services/launcher/launcher.yaml
xudifsd: Yes.. I also think so.

src/docker-images/RestfulAPI/runScheduler.sh
xudifsd: No, there are no such file in image, I think this is used for previous deployment when they tried to debug with local change mount directly into this directory. But we don't do this anymore.

PR #847, opened 2020-02-20
repair manager - configuration improvements

debbie-alaine:
-- added deployrepairmanagerconfig() method to deploy.py, which renders configs and copies them to persistent storage in infra node (etc/RepairManager/config). Docker build no longer required for configuration change.

-- simplified some duplicate config values (email authentication, cluster name)

-- added default recipient for email messages (set to DRI in config). If message['To'] is specified, default recipient (DRI) will be CC'd for all messages.


src/ClusterBootstrap/template/RepairManager/rule-config.yaml
```diff
@@ -7,11 +7,11 @@ rules:
...
     class_name: ECCRebootNodeRule
 
-rule_wait_time: 10 # time to sleep between rule execution (seconds)
+rule_wait_time: 30 # time to sleep between rule execution (seconds)
 
-cluster_name: {{cnf['repair-manager']['cluster_name']}}
+cluster_name: {{cnf['cluster_name_friendly']}}
 
-domain_name: {{cnf['repair-manager']['domain_name']}}
+portal_url: {{cnf['repair-manager']['portal_url']}}
```
Anbang-Hu: I suggest changing to a different name later on. `portal_url` usually indicates url including `https://`?

PR #849, opened 2020-02-21
old pipeline breaks fixed and doc updated

YinYangOfDao:
fix old pipeline breaks and update docs for old pipeline, tested for CPU clusters.


docs/deployment/Azure/cloud_init_configure.md
```diff
@@ -111,18 +111,37 @@ domain-offset:
   <can also set '*'>: <value0>
 
 repair-manager:
+  cluster_name: <the unique cluster name>
+  domain_name: <a domain name, e.g. dltshub.mydomain.com>
```
Anbang-Hu: `domain_name` was changed to `portal_url` in https://github.com/microsoft/DLWorkspace/pull/847/files#diff-dc066a8cecf60f7b4d06a012b30c3349R14

Anbang-Hu: I believe the entire section of `repair-manager` can be simplified. Please sync with @debbie-alaine

src/ClusterBootstrap/az_tools.py
```diff
@@ -921,6 +921,10 @@ def gen_cluster_config(output_file_name, output_file=True, no_az=False):
 
     if "sku_mapping" in config:
         cc["sku_mapping"] = config["sku_mapping"]
+        for sku in cc["worker_sku_cnt"]:
+            # this means that the cluster deployed with this pipeline cannot be heterogeneous
+            cc["gpu_type"] = cc["sku_mapping"].get(sku, {}).get('gpu-type', "None")
+            break
```
Anbang-Hu: Why is here a `break`?

docs/deployment/Azure/configure.md
```diff
@@ -1,80 +1,172 @@
...
+
+* infra_node_num: Should be odd (1, 3 or 5), number of infrastructure node for the deployment. 3 infrastructure nodes tolerate 1 failure, and 5 infrastructure nodes tolerate 2 failures. However, more infrastructure nodes (and more failure tolerance) will reduce performance of the node. 
+
+* worker_node_num: Number of worker node used for deployment. 
+
+* vm_image: Used to fix the image version if the changing LTS is breaking the consistency of the deployment.
+
+* nfs_vm: each item identified by `suffix` specs would describe an NFS node, and this item would overwrite default NFS specs. A `server_suffix` entry in `nfs_mnt_setup` should map to this item.
+
+* azure_location:
```
Anbang-Hu: No explanation for `azure_location`

docs/deployment/Azure/configure.md
```diff
@@ -1,80 +1,172 @@
...
+* azure_location: 
+
+Please use the following to find all available azure locations. 
+```
+az account list-locations
+```
+
+* infra_vm_size, worker_vm_size: infrastructure and worker VM size. 
+
+Usually, a CPU VM will be used for infra_vm_size, and a GPU VM will be used for worker_vm_size. Please use the following to find all available Azure VM size.
```
Anbang-Hu: "... find all available Azure VM size in a specific region, e.g. West US 2 in the below command"

docs/deployment/Azure/cloud_init_readme.md
```diff
@@ -32,11 +32,44 @@ First we need to setup the devbox that we use to operate on.
...
+
+to enter the mysql docker container.
+
+Enter the mysql container, run (-u -p according to your own setting in config):
+`mysql -u <user name> -<password>`
+
+Enter mysql, use `show databases` to list db, and run:
+`use <DLWS-initiated db>;`
+
+Add following entries for users:
```
Anbang-Hu: We should provide Admin commands to view and update ACL related fields in DB. But not in this PR.

docs/deployment/Azure/cloud_init_readme.md
```diff
@@ -32,11 +32,44 @@ First we need to setup the devbox that we use to operate on.
...
+
+INSERT INTO \`acl\` (\`id\`, \`identityName\`, \`identityId\`, \`resource\`, \`permissions\`, \`isDeny\`) VALUES (2, '<user account>', <uid>, 'Cluster', <3 for general users, 7 for cluster managers who is maintaining the cluster>, 0);
+
+INSERT INTO \`identity\` (\`id\`, \`identityName\`, \`uid\`, \`gid\`, \`groups\`) VALUES (2, '<user account>', <uid>, <gid>, < group info, e.g. `"[\"CCSAdmins\", \"MicrosoftUsers\"]"`);
+
+Then after existing the mysql docker container, we need to refresh the RestfulAPI cache:
```
Anbang-Hu: FYI, when we have an Admin tool for updating the ACL DB (via Restful API), we no longer need to restart Restful API server in this way.

docs/deployment/Azure/cloud_init_readme.md
Anbang-Hu: Theoretically, we only need to add to `acl` table. `user-synchronizer` will update/insert user info to `identity` table every hour.

src/ClusterBootstrap/az_tools.py
YinYangOfDao: because old pipeline is based on the assumption that the cluster is homogeneous. so there could be only one gpu_type

docs/deployment/Azure/cloud_init_configure.md
YinYangOfDao: depends on whether you want to allow cross-cluster repair-manager and whether we assume that alert-manager and repair-manager always use the same log in info etc.

PR #850, opened 2020-02-21
add job spec doc

xudifsd:
I tried to make this as complete as possible. If you find anything missing, please comment.

Our current launcher doesn't support arbitrary job roles, so I will add some check in restfulapi when implementing this feature, and return not implemented error when check failed, but the format should be like this.

### DIFF
--- a/docs/Jobs/spec.md
+++ b/docs/Jobs/spec.md
@@ -1,0 +1,63 @@
+# Job spec to  Restfulapi
+
+We use yaml format because it allow comments will note each field if it is required or optional, and gives example whenever possible.
+The yaml format is only for demostration purpose, we use json actually, so all keys are snake case.
+All fields are string unless specified otherwise.
+
+
+```yaml
+version: 2 # required, int
+user:
+  email: xxx@microsoft.com # required
+  vc_name: xxx # required
+job:
+  name: 'some name or even description' # required. At most 1024 characters
+  options: # optional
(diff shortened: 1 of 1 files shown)


docs/Jobs/spec.md
```diff
@@ -0,0 +1,63 @@
+# Job spec to  Restfulapi
+
+We use yaml format because it allow comments will note each field if it is required or optional, and gives example whenever possible.
```
Anbang-Hu: Change to 
"We use yaml format because it allows commenting each field, e.g. if it is required or optional, and examples whenever possible"

docs/Jobs/spec.md
```diff
@@ -0,0 +1,63 @@
...
+        min_succeeded_task_count: 1 # optional, int, can only be 1 currently. Consider this job failed when num of task from this role failed
+      resource:
+        require: # these will be set as resource limitation used by docker container
+          cpu: 200 # optional, default to cluster config
+          memory_mb: 3000 # optional, int, default to cluster config
+          gpu: 4 # optional, int, default to 0
+      cmd: text_that_be_interpreted_by_bash # required
+```
+
+Restfulapi will validate each field, and return 400 http code when validation failed. Will return uuid if job submission succeed.
```
Anbang-Hu: "Restfulapi will validate each field, and return 400 http code when validation fails. Restfulapi returns uuid if job submission is successful."

PR #851, opened 2020-02-22
update mysql fix tool to fit resourceQuota schema

YinYangOfDao:
support updating DB in deploy.py, support ratio-based auto-resource-assignment


src/ClusterBootstrap/deploy.py
```diff
@@ -3408,6 +3408,12 @@ def run_command(args, command, nargs, parser):
         if nargs[0] == "config":
             update_config_nodes()
 
+    elif command == "updatemysqlvc":
+        from MySQLDataHandler import update_db
+        with open("deploy/RestfulAPI/config.yaml") as f:
```
Anbang-Hu: Why loading `config.yaml` again? All configurations should already be in `config.yaml`?

YinYangOfDao: no, it's different yaml file. this MysqlDatahandler.py reads the config in RestfulAPI container, as in the code. deploy.py instead only reads ClusterBootstrap/config.yaml and cluster.yaml, if we don't read the one in the code, there would be key error

src/utils/MySQLDataHandler.py
```diff
@@ -247,6 +353,8 @@ def CreateTable(self):
             # when the VC has vm of same GPU type but different VMsizes, e.g., when VC has Standard_NC6s_v3 and Standard_NC12s_v3 both?
             # impossible since there's no way to do it with current config mechanism
 
+            vc_table_exist = table_exists(self.conn, self.vctablename)
```
Anbang-Hu: Why not make `table_exists` a class method?

src/utils/MySQLDataHandler.py
```diff
@@ -260,17 +368,19 @@ def CreateTable(self):
...
-            sql += default_vc_value
-
             cursor = self.conn.cursor()
             cursor.execute(sql)
             self.conn.commit()
             cursor.close()
 
+            if not vc_table_exist:
+                defalult_vc_rows = insert_vc_sqls(config)
+                execute_mysql(self.conn, defalult_vc_rows)
```
Anbang-Hu: There is a class method called `AddVC`

src/utils/MySQLDataHandler.py
```diff
@@ -43,10 +43,102 @@ def base64decode(str_val):
...
+    return sqls
+
+
+def column_exists(conn, tablename, columnname):
+    sql = "SHOW COLUMNS FROM `{}` LIKE '{}';".format(tablename, columnname)
+    cursor = conn.cursor()
+    cursor.execute(sql)
+    result = cursor.fetchone()
+    print(result)
+    return not result is None
```
Anbang-Hu: `result is not None`

src/utils/MySQLDataHandler.py
```diff
@@ -43,10 +43,102 @@ def base64decode(str_val):
...
+    total = 0
+    for entry in myresult:
+        total += sum(eval(entry[4]).values())
+    ratio = {}
+    for entry in myresult:
+        ratio[str(entry[2])] = sum(eval(entry[4]).values()) / total
+    return ratio
+
+
+def update_db(config):
```
Anbang-Hu: For update VC table, it's better to use `UpdateVC`

src/utils/MySQLDataHandler.py
```diff
@@ -330,6 +274,23 @@ def CreateTable(self):
             self.conn.commit()
             cursor.close()
 
+    @record
+    def column_exists(self, tablename, columnname):
+        sql = "SHOW COLUMNS FROM `{}` LIKE '{}';".format(tablename, columnname)
+        cursor = self.conn.cursor()
+        cursor.execute(sql)
+        result = cursor.fetchone()
+        print(result)
```
Anbang-Hu: Remove print. If you really want to print, use `logger`

Anbang-Hu: Same for line 292, 367, 373

src/utils/MySQLDataHandler.py
```diff
@@ -393,13 +354,25 @@ def UpdateStorage(self, vcName, url, storageType, metadata, defaultMountPath):
             logger.error('Exception: %s', str(e))
             return False
 
-
     @record
```
Anbang-Hu: Please move `@record` annotation to where it belonged - above `AddVC`

src/ClusterBootstrap/step_by_step.sh
```diff
@@ -21,7 +21,6 @@ fi
 ./deploy.py labelsku
 ./deploy.py webui
 ./deploy.py docker push restfulapi
-./deploy.py docker push webui
```
Anbang-Hu: Why removing this line?

src/utils/MySQLDataHandler.py
```diff
@@ -393,13 +354,25 @@ def UpdateStorage(self, vcName, url, storageType, metadata, defaultMountPath):
...
-    def AddVC(self, vcName, quota, metadata):
+    def init_vc_sqls(self, config):
+        ratio_dict = {config['defalt_virtual_cluster_name']: 1.0}
+        if "vc_resource_ratio" in config:
+            ratio_dict = config["vc_resource_ratio"]
+        quota, metadata, res_quota, res_meta = vc_value_str(config, ratio_dict)
+        if quota == "":
+            return
+        sqls = []
```
Anbang-Hu: What is `sqls = []` for?

src/utils/MySQLDataHandler.py
```diff
@@ -393,13 +354,25 @@ def UpdateStorage(self, vcName, url, storageType, metadata, defaultMountPath):
...
+        for i, (vc, vc_res_quota) in enumerate(res_quota.items()):
```
Anbang-Hu: `i` is not used. The for loop can be simplified to `for vc, vc_res_quota in res_quota.items():`

src/utils/vc_quota.py
```diff
@@ -0,0 +1,60 @@
...
+    res_meta_dict = {}
+    for r_type in ["cpu", "memory", "gpu", "gpu_memory"]:
+        tmp_res_meta = {}
+        for sku in worker_sku_cnt:
+            sku_name_in_map = sku if sku in sku_mapping else ""
+            pernode_cnt = sku_mapping.get(sku_name_in_map, {}).get(r_type, 0)
+            if "memory" in r_type:
+                pernode_cnt = '{}Gi'.format(pernode_cnt)
+            tmp_res_meta[sku_name_in_map] = {"per_node": pernode_cnt}
+            if r_type == "gpu":
```
Anbang-Hu: Need to add `unschedulable_ratio: 0.9` for CPU and memory. Check dev1's `vc` table.

src/ClusterBootstrap/az_tools.py
```diff
@@ -1212,9 +1212,6 @@ def run_command(args, command, nargs, parser):
 
     config = update_config(config)
```
Anbang-Hu: Can you revert the change in `az_tools.py`? It was checked in in #875

src/utils/MySQLDataHandler.py
Anbang-Hu: Please remove `@record` here. `init_vc_sqls` internally calls `AddVC`. It is not a direct DB call itself.

PR #853, opened 2020-02-22
Generalize VC resource

Anbang-Hu:
1. Expose CPU, memory, GPU consistently with `sku` label in VC view
2. Move VC quota calculation to `node_manager`
3. Use compact `json.dumps` for `ClusterStatus`
4. Fix `gpuLimit` unpopulated for distributed jobs when using `controller`

TODO:
1. A new process in `cluster_manager` to `DELETE FROM clusterstatus WHERE time < (NOW() - INTERVAL 1 DAY)`
2. Add default limit for CPU and memory for each job (RegularJob, PSDistJob, InferenceJob)


src/ClusterManager/virtual_cluster_status.py
```diff
@@ -0,0 +1,205 @@
...
+
+        return vc_statuses
+
+    def __get_vc_pod_statuses(self):
+        pod_statuses = self.cluster_status.pod_statuses
+        vc_pod_statuses = {}
+        for vc_name in self.vc_info:
+            vc_pod_statuses[vc_name] = {
+                name: pod_status
+                for name, pod_status in pod_statuses.items()
```
xudifsd: Maybe put this for loop to the outer loop? This essentially traverse the pod_statuses len(vc_inf) times

src/ClusterManager/virtual_cluster_status.py
```diff
@@ -0,0 +1,205 @@
...
+
+def get_vc_info(vc_list):
+    vc_info = {}
+    for vc in vc_list:
+        resource_quota = {}
+        try:
+            resource_quota = json.loads(vc["resourceQuota"])
+        except:
+            logger.exception("Parsing resourceQuota failed for %s", vc,
+                             exc_info=True)
```
xudifsd: I think exception do not need exec_info=True argument, it's by default in exception method

PR #858, opened 2020-02-24
Add DB manager

Anbang-Hu:
The size of `clusterstatus` is going wild since it's not GC'ed. 
Cluster status snapshots older than 1 day will be deleted. A minimum of 10 snapshots are to be maintained in the table. In theory, we only need to keep the last snapshot of cluster status in this table.

### DIFF
--- a/src/ClusterManager/db_manager.py
+++ b/src/ClusterManager/db_manager.py
@@ -44,21 +44,18 @@ def delete_old_cluster_status():
         if num_rows <= 10:  # Retain 10 rows for safety
             return
 
-        logger.info("Deleting rows from table %s older than %s", table,
+        logger.info("Deleting rows from table %s older than %s day(s)", table,
                     DAYS_AGO)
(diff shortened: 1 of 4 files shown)


src/ClusterManager/db_manager.py
```diff
@@ -0,0 +1,78 @@
...
+    while True:
+        update_file_modification_time("db_manager")
+
+        with manager_iteration_histogram.labels("db_manager").time():
+            try:
+                delete_old_cluster_status()
+            except:
+                logger.exception("Deleting old cluster status failed",
+                                 exc_info=True)
+        time.sleep(600)
```
hongzhili: This can be a daily job, say sleep 86400. Alternative is to create a new cluster garbage collection service, a k8s cron job. -- to clean DB as well as check the orphan pods in k8s.

Anbang-Hu: Let's use 86400 here and create a cron job service in a later PR.

PR #861, opened 2020-02-25
Refine jobs v2

Gerhut:
image
image
image

- Reorganize jobs into "active" ones and "inactive" ones in both my jobs and all jobs
  - both request 20 inactive jobs initially and will extend when checking the last page.
  - If you still want job id in the row, tell me and I will add it back
- JobId -> JobLink, moved into detail panel
- Better priority text input
- Move accurate submitted / started / finished time into detail panel, keep relative time in rows
  - submission time of active jobs
  - finished time of inactive jobs (But still ordered by submitted time in back-end)


Anbang-Hu:
Can we add job ID in the dropdown for each job in View Jobs page? It already appears in job link, but it would be easier to copy as a separate field.

PR #864, opened 2020-02-25
Add default resource for jobs of different types

Anbang-Hu:
This PR 
- eliminates the need to have special CPU configuration for a CPU cluster. 
- generalizes all resource (GPU, CPU, memory) assignment with sku node label.
- sets default resource values for both requests and limits for different job types

Upon adding `enable_job_resource_limit: True` in `config.yaml`, jobs now have performance isolation for both CPU and memory.

The default policy for filling in requests and limits for cpu and memory is
`sys_default = (cpu_request, cpu_limit, memory_request, memory_limit) = (1, 1, 2Gi, 2.5Gi)`

- `RegularJob`
    - gpu job: proportionally assign cpu and memory based on gpu request
    - cpu job: assign `sys_default`

- `PSDistJob`
    - gpu job: proportionally assign cpu and memory based on gpu request
    - cpu job: assign entire node if targeting cpu nodes; assign `sys_default` if targeting gpu nodes

- `InferenceJob`
    - gpu_job: proportionally assign cpu and memory based on gpu request
    - cpu job: assign `sys_default`

In current dev cluster, default values for different jobs:
```
# RegularJob 0 GPU
    resources:
      limits:
        cpu: "1"
        memory: 2560Mi
        nvidia.com/gpu: "0"
      requests:
        cpu: "1"
        memory: 2Gi
        nvidia.com/gpu: "0"

# RegularJob 1 GPU
    resources:
      limits:
        cpu: "6"
        memory: 112Gi
        nvidia.com/gpu: "1"
      requests:
        cpu: "5"
        memory: 103219Mi
        nvidia.com/gpu: "1"

# PSDistJob 1 node (4 GPU)
# ps-0
    resources:
      limits:
        cpu: "1"
        memory: 2Gi
        nvidia.com/gpu: "0"
      requests:
        cpu: "1"
        memory: "0"
        nvidia.com/gpu: "0"

# worker-0
    resources:
      limits:
        cpu: "24"
        memory: 448Gi
        nvidia.com/gpu: "4"
      requests:
        cpu: "21"
        memory: 412876Mi
        nvidia.com/gpu: "4"
```


src/ClusterBootstrap/template/RestfulAPI/config.yaml
```diff
@@ -86,3 +86,4 @@ elasticsearch:
   {% endfor %}
 
 job_mountpoints: {{cnf["job_mountpoints"]}}
+enable_job_resource: {{cnf["enable_job_resource"]}}
```
xudifsd: Maybe change name to `enable_job_resource_limit`? The name is not clear to me at first glance.

src/ClusterManager/pod_template.py
```diff
@@ -93,8 +93,6 @@ def generate_params(self, job):
 
         if "nodeSelector" not in params:
             params["nodeSelector"] = {}
-        if "gpuType" in params:
-            params["nodeSelector"]["gpuType"] = params["gpuType"]
         if "sku" in params:
```
xudifsd: I also see seom gpuType keywords in framework.py, maybe delete them as well.

src/utils/job_params_util.py
```diff
@@ -61,3 +79,276 @@ def get_resource_params_from_job_params(params):
         "gpu": gpu.to_dict(),
         "gpu_memory": gpu_memory.to_dict(),
     }
+
+
+class JobParams(object):
+    subclasses = {}
+
+    @classmethod
+    def register_subclass(cls, job_type):
```
xudifsd: It looks strange to me, since we do not need to be able to register subclass at runtime, why we need this method instead of just put a `subclasses = {"RegularJob": RegularJobParams ...}`?

src/utils/job_params_util.py
```diff
@@ -61,3 +79,276 @@ def get_resource_params_from_job_params(params):
...
+
+    def gen_sku(self):
+        sku = self.params.get("sku", "")
+        if sku != "":
+            self.sku = sku
+            return
+
+        sku_list = self.get_sku_list()
+        if len(sku_list) > 0:
+            self.sku = sku_list[0]
```
xudifsd: Always choose the first one?

src/utils/job_params_util.py
```diff
@@ -61,3 +79,276 @@ def get_resource_params_from_job_params(params):
...
+        else:
+            # For a cpu job, assign system default.
+            request, limit = self.get_default_cpu_request_and_limit()
+        return request, limit
+
+    @override
+    def get_memory_request_and_limit(self):
+        if self.gpu_limit > 0:
+            # For a gpu job, proportionally assign memory according to gpu.
+            request, limit = self.get_memory_proportional()
```
xudifsd: I still have some concerns here. User do not know how much memory they can get if they didn't specify sku and memory request/limit. Maybe we need to expose environment variables about the resource. I'm not sure how users thinking about this.

src/ClusterManager/pod_template.py
Anbang-Hu: Those are in `labels`. Let's do this in another PR.

src/utils/job_params_util.py
Anbang-Hu: We can define a configurable policy for this when we do global scheduling. For now, it's sufficient.

src/utils/job_params_util.py
Anbang-Hu: Let's discuss and resolve this concern in a future PR.

PR #875, opened 2020-02-27
Fix deployment break for elasticsearch and nfs

Anbang-Hu:


src/ClusterBootstrap/deploy.py
```diff
@@ -1544,10 +1547,13 @@ def update_elasticsearch_nodes(nargs):
     TODO: Should be covered by update_role_nodes in deploy.py V2
     """
     # This is to temporarily replace gpu_type with None to disallow nvidia runtime config to appear in /etc/docker/daemon.json
-    prev_gpu_type = config["gpu_type"]
+    prev_gpu_type = config.get("gpu_type")
     config["gpu_type"] = "None"
     utils.render_template_directory("./template/kubelet", "./deploy/kubelet", config)
-    config["gpu_type"] = prev_gpu_type
+    if prev_gpu_type is None:
+        config.pop("gpu_type", None)
```
YinYangOfDao: kubelet/daemon.json use this condition, {% if "gpu_type" in cnf and cnf["gpu_type"] != "None" %}
maybe populate with the str "None"?

src/ClusterBootstrap/deploy.py
```diff
@@ -1477,11 +1477,14 @@ def update_nfs_nodes(nargs):
...
     """
     # This is to temporarily replace gpu_type with None to disallow nvidia runtime config to appear in /etc/docker/daemon.json
-    prev_gpu_type = config["gpu_type"]
+    prev_gpu_type = config.get("gpu_type")
     config["gpu_type"] = "None"
     utils.render_template_directory(
         "./template/kubelet", "./deploy/kubelet", config)
-    config["gpu_type"] = prev_gpu_type
+    if prev_gpu_type is None:
+        config.pop("gpu_type", None)
```
YinYangOfDao: would this overwrite line 1481?

Anbang-Hu: Yes. `prev_gpu_type is None` indicates that `"gpu_type"` did not exist. We need to remove it to restore `config`.

PR #876, opened 2020-02-27
repairmanager refactoring - Actions

debbie-alaine:
DO NOT MERGE, not fully tested.

Opening PR for initial feedback on refactoring.

-- Created an Action object that can be reused in future repair manager rules/policies


src/RepairManager/actions/cordon_action.py
```diff
@@ -0,0 +1,22 @@
+import os, sys
+sys.path.append(os.path.dirname(os.path.abspath(__file__)))
```
Anbang-Hu: I don't think `sys.path.append` is need if the files to import are in the same directory

PR #880, opened 2020-02-28
Revert "Do not write log to job directory. (#841)"

Gerhut:
This reverts commit 4815f87d9d022292f1ad17d3a982ebb260f43435.

### DIFF
--- a/src/ClusterManager/job_manager.py
+++ b/src/ClusterManager/job_manager.py
@@ -25,7 +25,7 @@ sys.path.append(
     os.path.join(os.path.dirname(os.path.abspath(__file__)), "../utils"))
 
 from DataHandler import DataHandler
-from config import config
+from config import config, GetStoragePath
 import notify
 import k8sUtils
 from cluster_resource import ClusterResource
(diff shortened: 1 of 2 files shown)


src/ClusterManager/job_manager.py
```diff
@@ -333,12 +333,24 @@ def UpdateJobStatus(redis_conn,
...
+                                                 jobParams["dataPath"])
+    localJobPath = os.path.join(config["storage-mount-path"], jobPath)
+    logPath = os.path.join(localJobPath, "logs/joblog.txt")
+
+    if "userId" not in jobParams:
+        jobParams["userId"] = "0"
+
     if result == "Succeeded":
-        joblog_manager.extract_job_log(job["jobId"])
+        joblog_manager.extract_job_log(job["jobId"], logPath,
```
Anbang-Hu: This means writing to NFS upon success/failure even if elasticsearch is deployed. @xudifsd do you think it's ok?

xudifsd: This will be temporary, right?

Anbang-Hu: No, this will save to `/dlwsdata/work/<alias>/jobs/...../logs/joblog.txt`

xudifsd: I know, I meant this solution will be temporary, right? Then I'm fine with this.

PR #887, opened 2020-03-03
move public/private ssh key into DB

xudifsd:
This PR move generation of public/private key from user_manager to user_synchronizer. Also DB becomes a primary location for storing key pairs, and user_manager will dump key pairs from DB to NFS for backward compatible reason.

Step of deployment:

1. stop user_synchronizer/restfulapi/jobmanager in order (both local and global restfulapi)
2. alter table and dump data use script
    1. scp script to infra node
    2. exec `sudo pip3 install mysql-connector-python` in infra node
    3. exec `sudo /tmp/move_keys_into_db.py alter` in infra node
    4. exec `sudo /tmp/move_keys_into_db.py dump` in infra node
3. deploy jobmanager/restfulapi/user_synchronizer in order


xudifsd:
@Gerhut Please help review changes related to user-synchronizer. I have already tested in the real cluster.

xudifsd:
@Anbang-Hu Ready for review now.

src/user-synchronizer/__main__.py
```diff
@@ -17,6 +17,7 @@
 
 # Read environment variables
 domain_offset_file = environ.get('DOMAIN_OFFSET_FILE', None)
+restfulapi_url = environ['RESTFULAPI_URL']
```
Gerhut: I didn't find any more uses of `restfulapi_url` in this file except calls in `restfulapi.py`.

If this is just a restructure of moving all env vars to `__main__` instead of the actual module, could you please help move env retrieving in `logger.py` and `microsoft_graph.py` to `__main__` as well? This will prevent confusion in future code readers.

xudifsd: I moved to __main__ because I want to be able to test restfulapi.py by calling it directly (added __name__ check there).

Ok, I think environment variable read should happen in main module, not library. I will move variables from those module to main as well.

xudifsd: I'm a little confused about Fetch token part in `microsoft_graph.py`, the fetched token seems not used by any function. oauth will be configured by calling `fetch_token`?

src/init-scripts/setup_ssh_config.sh
```diff
@@ -97,24 +94,38 @@ DLTS_ROLE_IDX
...
-chown ${DLTS_USER_NAME} ${AUTHORIZED_FILE}
-chmod 600 ${AUTHORIZED_FILE}
 
-set -x
+PRIVATE_KEY_FILE=/home/${DLTS_USER_NAME}/.ssh/id_rsa
+printenv DLTS_SSH_PRIVATE_KEY > $PRIVATE_KEY_FILE
+chown ${DLTS_USER_NAME} ${SSH_ENVIRONMENT_FILE} ${AUTHORIZED_FILE} ${PRIVATE_KEY_FILE}
+chmod 600 ${SSH_ENVIRONMENT_FILE} ${AUTHORIZED_FILE} ${PRIVATE_KEY_FILE}
+
+PRIVATE_KEY_FILE=/home/${DLTS_USER_NAME}/.ssh/id_rsa
```
Anbang-Hu: Line 124-127 is a dup of line 119 - 122

src/user-synchronizer/__main__.py
Gerhut: Yes, it will configure `self.token` as well as return it at the same time.

PR #888, opened 2020-03-03
Use jobLogCursor column to store log cursors.

Gerhut:
For current clusteres, add a column to existing table `jobs`:
```
ALTER TABLE ADD COLUMN jobLogCursor LONGTEXT  NULL;
```


Anbang-Hu:
We will need to log onto NFS as a fallback even if ES is enabled.

Gerhut:
Added temporary `__extract_job_log_legacy` option to forcely tell jobmanager to use the kubectl command to retrieve logs. Tested by looking at the database `jobLog` field of newly created jobs.

PR #894, opened 2020-03-05
fix default elastic search label, improve config rendering for devs

YinYangOfDao:


YinYangOfDao:
to add elastic search with cloudinit:
0. set up 1 extra worker.
kubectl label nodes <the node as elasticsearch node> elasticsearch=active

PR #900, opened 2020-03-08
Cloudinit update

YinYangOfDao:
updated following items:

1. allow infra to access NFS port 10250,
2. add restricted source address prefixes, 
3. fix cpu runtime error
4. use private registry, 
5. add admin tool to quickly stop/start/render service, or update storagemanager/restfulapi config
6. fix ci verify deployment
7. change permission of sshkey from 700 to 600
8. add download counterpart to download kubectl 

there is a special case that rarely happen, but if you want to put all service level docker to a private registry, please specify "private_docker_registry_username" and "private_docker_registry_password" in config.


YinYangOfDao:
elastic search not well fixed. would need to refactor template rendering part, cannot finish by Monday.

PR #909, opened 2020-03-09
Add samba as a default role

Anbang-Hu:
Allow execution of command `./ctl.py connect samba 0`.


Anbang-Hu:
@YinYangOfDao can you help check if this change will break any part of the (cloud init) deployment?

PR #914, opened 2020-03-10
change how network metric scrapped

xudifsd:
Previously we use `iftop` to get all connections in host and traffic in these connections, and then we distributed traffic to jobs with knowledge of which connection belongs to which container. We gained such knowledge by using `lsof` in each container namespace.

Such method may miss some data when connection is short and doesn't fetched by `iftop`, so the traffic is out of record.

With this PR. We get network metric using following method instead:
* If container is using host network, we assume this container take all resource of the host, hence we get its network consumption from host's real ethernet.
* else we get network consumption by first get virtual ethernet using techniq from [here](https://www.digitalocean.com/community/tutorials/how-to-inspect-kubernetes-networking#finding-a-pod%E2%80%99s-virtual-ethernet-interface), and get network consumption of this virtual ethernet.

Hopefully with this method, network metrics of both host network container and non host network container are more accurate.

This also changed meaning of `task_net_in_byte` and `task_net_out_byte`, previously these two metrics represent byte rate, now they represent byte sent/received.

### DIFF
--- a/src/ClusterBootstrap/services/monitor/grafana-config/job-status-dashboard.json
+++ b/src/ClusterBootstrap/services/monitor/grafana-config/job-status-dashboard.json
@@ -382,7 +382,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "task_net_in_byte{job_name=\"$job_name\"}",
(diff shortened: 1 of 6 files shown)


Anbang-Hu:
there is a test failure

xudifsd:
> there is a test failure

yes, will fix this today.

PR #917, opened 2020-03-10
.ssh folder needs to have 700

Anbang-Hu:
`.ssh` needs to have 700, otherwise, samba access fails.


xudifsd:
maybe add a comment?

Anbang-Hu:
> maybe add a comment?

Added

PR #927, opened 2020-03-11
fix init container build break

xudifsd:
install wget with `--no-install-recommends` will not install let's encrypt CA, and zlib.net uses it. Hence break the ssh build in init container.


xudifsd:
related to #905

PR #934, opened 2020-03-16
Whitelist user ip in cloud init

Anbang-Hu:


src/ClusterBootstrap/cloud_init_aztools.py
```diff
@@ -565,6 +566,91 @@ def get_deployed_cluster_info(config, args):
...
+        for rule in rules:
+            if rule.get("name") == "whitelist":
+                return rule.get("sourceAddressPrefixes", [])
+    except Exception as e:
+        print("Exception: %s" % e)
+
+    return []
+
+
+def add_nsg_rule_whitelist(ips, dry_run=False):
```
YinYangOfDao: is it possible to encounter overlapping ranges? safer to use utils.keep_widest_subnet to pre-process, unless we can assume that the whitelist args are always valid.
also, users might need to know that input IPs are concatenated by comma

PR #939, opened 2020-03-17
smooth elasticsearch and other role machines

YinYangOfDao:
support elasticsearch node, and making it more natural toward supporting more generally, all nodes that are of other roles than worker and are homogeneous

### DIFF
--- a/src/ClusterBootstrap/cloud_init_aztools.py
+++ b/src/ClusterBootstrap/cloud_init_aztools.py
@@ -53,7 +53,7 @@ def update_config_resgrp(config):
 
 def load_sshkey(config):
     assert os.path.exists(
-        './deploy/sshkey/id_rsa.pub') and "Generate SSHKey first!"
+        './deploy/sshkey/id_rsa.pub'), "Generate SSHKey first!"
     with open('./deploy/sshkey/id_rsa.pub') as f:
         config["azure_cluster"]["sshkey"] = f.read()
     return config
@@ -127,7 +127,7 @@ def create_vnet(config, args):
 
 
 def create_nsg(config, args):
-    assert "source_addresses_prefixes" in config["cloud_config_nsg_rules"]["dev_network"] and "Please \
+    assert "source_addresses_prefixes" in config["cloud_config_nsg_rules"]["dev_network"], "Please \
     setup source_addresses_prefixes in config.yaml, otherwise, your cluster cannot be accessed"
     source_addresses_prefixes = config["cloud_config_nsg_rules"][
         "dev_network"]["source_addresses_prefixes"]
@@ -202,7 +202,7 @@ def create_nsg(config, args):
 
 
 def create_nfs_nsg(config, args):
-    assert "source_addresses_prefixes" in config["cloud_config_nsg_rules"]["dev_network"] and "Please \
+    assert "source_addresses_prefixes" in config["cloud_config_nsg_rules"]["dev_network"], "Please \
     setup source_addresses_prefixes in config.yaml, otherwise, your cluster cannot be accessed"
     source_addresses_prefixes = config["cloud_config_nsg_rules"][
         "dev_network"]["source_addresses_prefixes"]
@@ -261,12 +261,12 @@ def deploy_cluster(config, args):
 
 
 def validate_machine_spec(config, spec):
-    assert "role" in spec and ((set(spec["role"]) - set(config["allroles"])) == set()) and \
+    assert "role" in spec and ((set(spec["role"]) - set(config["allroles"])) == set()), \
         "must specify valid role for vm!"
     if "name" in spec:
-        assert spec["number_of_instance"] <= 1 and "cannot overwirte name for multiple machines one time!"
+        assert spec["number_of_instance"] <= 1, "cannot overwirte name for multiple machines one time!"
     if "nfs" in spec["role"]:
-        assert spec["number_of_instance"] <= 1 and "NFS machine spec must be configured one by one!"
+        assert spec["number_of_instance"] <= 1, "NFS machine spec must be configured one by one!"
 
 
 def gen_machine_list_4_deploy_action(complementary_file_name, config):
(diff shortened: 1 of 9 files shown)


src/ClusterBootstrap/cloud_init_aztools.py
```diff
@@ -334,6 +334,9 @@ def add_machine_in_parallel(cmds, args):
 
 
 def add_machine(vmname, spec, verbose, dryrun, output_file):
+    multual_exclusive_roles = set(["infra", "worker", "elasticsearch", "mysqlserver"])
```
Anbang-Hu: We need to refactor his function in a separate PR. The logic has become somewhat untrackable.

src/ClusterBootstrap/cloud_init_aztools.py
```diff
@@ -334,6 +334,9 @@ def add_machine_in_parallel(cmds, args):
...
+    mul_ex_role_in_spec = list(set(spec["role"]) & multual_exclusive_roles)
+    assert len(mul_ex_role_in_spec) <= 1 and "We don't allow role overlapping between these roles."
```
Anbang-Hu: Can we use `assert <condition>, <msg>` instead of `assert <condition> and <msg>`? Please include what roles are not allowed overlapping in the message as well.

src/ClusterBootstrap/cloud_init_deploy.py
```diff
@@ -500,23 +500,29 @@ def add_implied_services_based_on_config(config):
...
         mounting = yaml.safe_load(mf)
     config["mount_and_link"] = []
     for mnt_itm in mounting.values():
         for fs in mnt_itm["fileshares"]:
             config["mount_and_link"] += fs["remote_mount_path"],
+    return config
 
+
+def escaped_etcd_end_point_and_k8s_api_server(config):
+    config["escaped_etcd_endpoints"] = ",".join(
```
Anbang-Hu: Can we stick with `etcd_endpoints` and `api_servers`? Why adding `escaped`?

src/ClusterBootstrap/cloud_init_deploy.py
```diff
@@ -557,13 +556,26 @@ def render_worker_node_specific(config, args):
                               "./deploy/cloud-config/cloud_init_worker_{}.txt".format(sku), config)
 
 
+def render_elasticsearch_node_specific(config, args):
+    if not "elasticsearch_node" in config or len(config["elasticsearch_node"]) == 0:
+        print("Warning: no elasticsearch node specified, logging service might not work as expected.")
+        return
+    config.pop("mount_and_link", [])
+    config = escaped_etcd_end_point_and_k8s_api_server(config)
+    example_node_name = config["elasticsearch_node"][0].split(".")[0]
```
Anbang-Hu: Naming `example_node_name` is not clear to me at first glance.

src/ClusterBootstrap/cloud_init_deploy.py
```diff
@@ -557,13 +556,26 @@ def render_worker_node_specific(config, args):
...
+    common_labels = get_kube_labels_of_machine_name(config, example_node_name)
+    default_f2cp = ["kubernetes_common", "kubelet_worker"]
+    config["kube_labels"] = common_labels
+    config["file_modules_2_copy"] = [mod for mod in default_f2cp]
```
Anbang-Hu: ` config["kube_labels"] = get_kube_labels_of_machine_name(config, example_node_name)`
`config["file_modules_2_copy"] = ["kubernetes_common", "kubelet_worker"]`?

src/ClusterBootstrap/scripts/cloud_init_worker.sh
```diff
@@ -15,5 +15,8 @@ source ../boot.env
 ./render_env_vars.sh worker.kubelet.service.template /etc/systemd/system/kubelet.service KUBE_LABELS
 
 bash ./post-worker-deploy.sh
+
+if [ ! -z "$MNT_N_LNK" ]; then
```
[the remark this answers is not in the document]
YinYangOfDao: no issue, just to avoid confusion in cloudinit-log

src/ClusterBootstrap/cloud_init_deploy.py
YinYangOfDao: to avoid pop-out-to-recover, and overriding issue when only certain commands are executed/re-executed.

src/ClusterBootstrap/cloud_init_deploy.py
YinYangOfDao: [mod for mod in default_f2cp] is a deepcopy, otherwise later the list would be overwriten, so this is used to avoid unexpected influence on other rendered files.

src/ClusterBootstrap/cloud_init_aztools.py
YinYangOfDao: sure, plz let me know the preferred way

src/ClusterBootstrap/cloud_init_deploy.py
YinYangOfDao: to extrapolate to other roles as well. can discuss this together in the separate PR mentioned above.

src/ClusterBootstrap/cloud_init_aztools.py
YinYangOfDao: corrected.

src/ClusterBootstrap/cloud_init_deploy.py
Anbang-Hu: Can you give an example of such issue?

src/ClusterBootstrap/cloud_init_deploy.py
Anbang-Hu: `first_node`? Since you are literally just taking the first element in the node list.

src/ClusterBootstrap/cloud_init_deploy.py
YinYangOfDao: renderspecific nfs after render worker etc.

src/ClusterBootstrap/cloud_init_deploy.py
YinYangOfDao: ok, I'll rename it to first_node in that refactor PR.

src/ClusterBootstrap/cloud_init_deploy.py
YinYangOfDao: actually previously there are more "\/" in the escaped url in NFS's cloudinit file before. Using an independent variable is safer, and make the diff between escaped url v.s. env var name clear

PR #946, opened 2020-03-19
NSG rules with service tag

Anbang-Hu:
To allow certain Azure resources to access DLTS resources, it is better to use Azure owned service tag. This way, we do not need to maintain a long IP range list in `restricted_source_address_prefixes`.

In `config.yaml`:
```
cloud_config_nsg_rules:
  service_tags:
  - ServiceTag1
  - ServiceTag2
```

`./az_tools.py service_tag_rules create`
`./az_tools.py service_tag_rules delete`

`./cloud_init_aztools.py service_tag_rules create`
`./cloud_init_aztools.py service_tag_rules delete`

### DIFF
--- a/src/ClusterBootstrap/az_tools.py
+++ b/src/ClusterBootstrap/az_tools.py
@@ -13,6 +13,8 @@ import yaml
 import utils
 from az_params import *
 from params import *
+from az_utils import create_nsg_rules_with_service_tags, \
+    delete_nsg_rules_with_service_tags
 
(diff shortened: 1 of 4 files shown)


Anbang-Hu:
> would rather have the refactoring part in cloud_init_aztools.py postponed. would cause conflict with my other 2 PRs.

I went through your PRs. There shouldn't be more than 3 conflicts. Please resolve it accordingly.

PR #948, opened 2020-03-19
Add a simple document website

Gerhut:
Preview on <http://d*****b-int:8090/>

- [ ] Resolve symlink problems: Docker won't include synlink files into its build context.
- [ ] Make more `##` titles for better search.
- [ ] Add CI to keep document format well
- [ ] Fix links


Anbang-Hu:
Thanks @Gerhut for initiating the wiki website! Two questions:
1. Can we enable AAD authentication?
2. What if we have docs only for internal usage?

Gerhut:
Switched to internal repo

PR #953, opened 2020-03-20
fix failed test case

xudifsd:
seems get log logical has been changed, previous will stop when cursor is None, current is should stop when cursor is equals to current one.

@Gerhut please confirm.


Gerhut:
Unfortunately this is a bug...

PR #956, opened 2020-03-21
easy admin: intoj

xudifsd:
For easy admin.

put `dlts_get_cmd ` and `intoj` in PATH, config `REPO_DIR` in `dlts_get_cmd` to directory where you put all DLWorkspace repo.

And then executing `intoj CLUSTER_NAME` anywhere will land you in the container of job manager of that cluster.

Will need `ctl.py` fix execute cmd without stdin attached problem.


xudifsd:
Ref #996

PR #957, opened 2020-03-21
Force entire node allocation and reject zero worker for distributed job

Anbang-Hu:
All integration tests passed with some of them ran multiple times due to slow log pulling.


src/utils/job_params_util.py
```diff
@@ -92,6 +92,9 @@ def __init__(self, params, quota, metadata, config):
         self.metadata = metadata
         self.config = config
 
+        # Allow partial node for worker for internal jobs.
+        self.internal = params.get("_internal", False)
```
hongzhili: We may need to check the permission in the future.

Anbang-Hu: added

src/ClusterManager/test/utils.py
```diff
@@ -120,6 +120,7 @@ def gen_default_job_description(
         "resourcegpu": 0,
         "memorylimit": "500M",
         "cpulimit": 1,
+        "_internal": True,
```
xudifsd: Maybe a more informative name?

PR #972, opened 2020-03-26
Logs can contain unicode

Anbang-Hu:


src/ClusterManager/joblog_manager.py
```diff
@@ -164,7 +164,7 @@ def extract_job_log_legacy(jobId, logPath, userId):
                     containerLogPath = os.path.join(
                         jobLogDir,
                         "log-container-" + log["containerID"] + ".txt")
-                    with open(containerLogPath, 'w') as f:
+                    with open(containerLogPath, 'w', encoding="utf-8") as f:
```
Anbang-Hu: @Gerhut shall we add this to the other logging branch (from ES/Azure Log Analyzer) as well?

Gerhut: I'll handle this

PR #978, opened 2020-03-27
rollback to old way of generating ssh key

xudifsd:
SInce putty do not support format python cryptography generated


xudifsd:
Ref #887

PR #979, opened 2020-03-29
Dashboard/error msg

Gerhut:
Dashboard will retrieve job status details when there is a "failed" state.


xudifsd:
What it looks like? Could you attach a screenshot when creating front end PR. I really do not know front end code.

Gerhut:
> What it looks like? Could you attach a screenshot when creating front end PR. I really do not know front end code.

Deployed to INT

xudifsd:
It's may because the jobmanager is not the up to dated one. I can see result now. Looks good.

image

PR #985, opened 2020-03-31
Bring job["log"] back in v1 API

Gerhut:
Ref: https://github.com/microsoft/DLWorkspace/pull/784/files#diff-1d886fc19ee0a084ef132f0fb1b00133L689-L712

Note: if there is no elasticsearch deployed or `__extract_job_log_legacy` set, jobmanager will write active job log to db and jobdetail v1 will response with logs, otherwise jobdetail will have `job["log"] = ""`


xudifsd:
We may need a test case for this. I will write one.

PR #989, opened 2020-03-31
kill etcd3 container if exist before start it in etcd_service.sh

YinYangOfDao:


src/ClusterBootstrap/template/etcd/etcd_service.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 /opt/resolve_ip.sh
 ETCDIP=$(cat /opt/defaultip)
+/usr/bin/docker rm -f etcd3 || true
```
xudifsd, line 4: Why need to have `|| true`?

Anbang-Hu: `true` can be removed.

PR #990, opened 2020-04-01
Dashboard: Refine tables in Jobs v2

Gerhut:
- [x] Switch between name and id
- [x] Sortable in most columns
- [x] Split active jobs into different status
    
image
image


Anbang-Hu:
Will job id work in the search box in View and Manage Jobs page?

Gerhut:
Yes, only v2 page

PR #993, opened 2020-04-01
use bash to execute inference job

xudifsd:
user will use source which doesn't exist in sh

### DIFF
--- a/src/ClusterManager/framework.py
+++ b/src/ClusterManager/framework.py
@@ -318,7 +318,7 @@ def gen_containers(job, role):
     if job.init_image is None:
         cmd = [
             "sh", "-c",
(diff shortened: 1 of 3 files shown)


src/ClusterManager/framework.py
```diff
@@ -318,7 +318,7 @@ def gen_containers(job, role):
     if job.init_image is None:
         cmd = [
             "sh", "-c",
-            "printenv DLWS_LAUNCH_CMD > /job_command.sh ; mkdir -p /dlts-runtime/status ; touch /dlts-runtime/status/READY ; sh /job_command.sh"
+            "printenv DLWS_LAUNCH_CMD > /job_command.sh ; mkdir -p /dlts-runtime/status ; touch /dlts-runtime/status/READY ; bash -x /job_command.sh"
```
Anbang-Hu, line 321: Should we change this for distributed and regular job as well?

xudifsd: solved

PR #999, opened 2020-04-02
expose more info to jobStatusDetail

xudifsd:
Added two test cases to make sure the behaviour is consistent.

### DIFF
--- a/src/ClusterManager/test/test_distributed_job.py
+++ b/src/ClusterManager/test/test_distributed_job.py
@@ -543,3 +543,28 @@ def test_fault_tolerance(args):
                                            cmd)
         assert code == 0, "code is %s, output is %s" % (code, output)
         assert output == "dummy\n", "output is %s" % (output)
+
+
+@utils.case()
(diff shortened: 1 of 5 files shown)


src/ClusterManager/test/test_distributed_job.py
```diff
@@ -543,3 +543,28 @@ def test_fault_tolerance(args):
                                            cmd)
         assert code == 0, "code is %s, output is %s" % (code, output)
         assert output == "dummy\n", "output is %s" % (output)
+
+
+@utils.case()
+def test_image_pull_msg(args):
+    expected = "ImagePullBackOff"
```
Anbang-Hu, line 550: We should fail this case at some point. Maybe in a separate PR.

xudifsd: Yes, this case is very hard to handle, because  when an image is huge, it will show this state until image is pulled successfully. We can have a timeout on this, but may fail on case of huge image.

PR #1000, opened 2020-04-03
Logging/deployment refinements

Gerhut:


src/docker-images/azure-blob-adapter/wsgi.py
```diff
@@ -24,37 +24,51 @@ def application(request):
     :param PlainRequest request:
     '''
     try:
-        blob_name = request.headers['X-Tag']
-    except KeyError:
-        logger.exception('Key Error')
-        return Response(status=400)
+        if request.method == 'GET' and request.path == '/healthz':
+            def get_container():
```
xudifsd, line 28: should define this function in top level. This actually define this method everytime liveness probe calls

src/docker-images/azure-blob-adapter/wsgi.py
```diff
@@ -24,37 +24,51 @@ def application(request):
...
-        logger.exception('Azure HTTP Error')
-        return Response(status=502)
+        if request.method == 'POST' and request.path == '/':
+            try:
+                blob_name = request.headers['X-Tag']
+            except KeyError:
+                logger.exception('Key Error')
+                return Response(status=400)
+
+            def append_blob():
```
xudifsd, line 44: Same here, define these methods in top level

src/docker-images/azure-blob-adapter/wsgi.py
```diff
@@ -24,37 +24,51 @@ def application(request):
...
+                resource_properties = append_blob_service.append_blob_from_bytes(
+                    container_name=container_name,
+                    blob_name=blob_name,
+                    blob=request.get_data(),
+                    count=request.content_length)
+                logger.info('Successfully append {} bytes to blob {} in container {}: {}'.format(
```
xudifsd, line 50: Do we really need such log in production?

Gerhut: You're right, since I enabled azure internal info logs, all of these info are directly outputted by azure library

src/docker-images/azure-blob-adapter/wsgi.py
Gerhut: since logs are removed, inlined these function calls

src/ClusterBootstrap/services/logging/fluent-bit.yaml
```diff
@@ -4,31 +4,41 @@ metadata:
...
+        prometheus.io/port: "{{cnf["fluent_bit"]["port"]}}"
         prometheus.io/path: /api/v1/metrics/prometheus
     spec:
       containers:
       - name: fluent-bit
         image: fluent/fluent-bit:1.4.1
         imagePullPolicy: Always
+        resources:
+          limits:
+            memory: 256Mi
```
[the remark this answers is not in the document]
Gerhut: Just copied from fluentd, will look it behaves and increase them.

PR #1002, opened 2020-04-03
Azure Blob Logging: Get last 1MB of log and sort manually

Gerhut:


src/utils/JobLogUtils.py
```diff
@@ -33,43 +33,34 @@ def GetJobLog(jobId, cursor=None, size=None):
...
-                    logger.exception('Failed to parse cursor')
-            end_range = start_range + CHUNK_SIZE
             lines = []
-            next_start_range = start_range
 
             try:
+                blob = append_blob_service.get_blob_properties(
+                    container_name=container_name,
+                    blob_name=blob_name)
+                start_range = max(0, blob.properties.content_length - CHUNK_SIZE)
```
xudifsd, line 42: Can this handle UTF-8 characters?

src/utils/JobLogUtils.py
```diff
@@ -33,43 +33,34 @@ def GetJobLog(jobId, cursor=None, size=None):
...
                 for i, chunk_line in enumerate(chunk_lines, 1):
                     try:
                         line = loads(chunk_line)
                         lines.append(line)
                     except JSONDecodeError:
-                        if i == len_chunk_lines:
-                            # Normal case, invalid JSON at the end of the log:
-                            #     Directly exit the loop and skip the increase of next_start_range
-                            break
+                        if i == 1:
```
xudifsd, line 54: Can you make it an independant function for parsing raw log? So we can write a UT for this. You should do it in the future.

src/utils/JobLogUtils.py
Gerhut: Manually ignored docode errors

PR #1003, opened 2020-04-03
make exit code distinct

xudifsd:
to facilitate debug


xudifsd:
This PR also add ssh self test after setup ssh config. And will fail with specific code when test failed. This will eliminate need for ps check for ssh connection. Hopefully will fail fast.

PR #1007, opened 2020-04-06
read/create sync configmap with version attached

xudifsd:
So that later retries will not read configmap of previous version


Anbang-Hu:
Will all the `configmaps` be cleaned up after the job is done?

xudifsd:
@Anbang-Hu Yes, job manager delete the job using label jobId=id, so will delete all of them

xudifsd:
but I do find there are many dead configmap not being cleaned. Will look into it when I have time.

PR #1011, opened 2020-04-07
refactor code

YinYangOfDao:
to change a specific service config of an already deployed node, modify status.yaml, not config.yaml.

### DIFF
--- a/src/ClusterBootstrap/cloud_init_aztools.py
+++ b/src/ClusterBootstrap/cloud_init_aztools.py
@@ -8,6 +8,7 @@ import argparse
 
 from az_params import *
 from params import *
+from constants import ENV_CNF_YAML, ACTION_YAML, STATUS_YAML
 from utils import random_str, keep_widest_subnet, \
     multiprocess_with_func_arg_tuples, exec_cmd_local, \
     execute_or_dump_locally
@@ -29,10 +30,6 @@ from ctl import run_kubectl
 sys.path.append("../utils")
 from ConfigUtils import add_configs_in_order, merge_config
 
-ENV_CNF_YAML = "config.yaml"
-STATUS_YAML = "status.yaml"
-ACTION_YAML = "action.yaml"
-
 
 def init_config():
     config = {}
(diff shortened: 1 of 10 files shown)


src/ClusterBootstrap/cloud_init_aztools.py
```diff
@@ -17,15 +17,22 @@
...
     delete_logging_storage_account, \
     get_connection_string_for_logging_storage_account
 
 from cloud_init_deploy import load_node_list_by_role_from_config
+from ctl import run_kubectl
 sys.path.append("../utils")
-from ConfigUtils import add_configs_in_order
+from ConfigUtils import add_configs_in_order, merge_config
+
+ENV_CNF_YAML = 'config.yaml'
```
Anbang-Hu, line 32: Can we consistently use double quotes for strings?

src/ClusterBootstrap/cloud_init_aztools.py
```diff
@@ -17,15 +17,22 @@
...
+STATUS_YAML = 'status.yaml'
+ACTION_YAML = 'action.yaml'
```
Anbang-Hu, line 34: If you replace `az_complementary.yaml` with `action.yaml`, you will need to replace it everywhere, especially `src/ClusterManager/test/utils.py`, otherwise @xudifsd won't be happy.

YinYangOfDao: sure, I would update all existing yaml files.

src/ClusterBootstrap/cloud_init_aztools.py
YinYangOfDao: corrected

src/ClusterBootstrap/cloud_init_aztools.py
```diff
@@ -772,4 +821,4 @@ def run_command(command, config, args):
     nargs = args.nargs
```
Anbang-Hu, line 821: I don't see any use of `nargs` if you choose to remove it in line 775

src/ClusterBootstrap/ctl.py
```diff
@@ -24,17 +24,19 @@
 from cloud_init_deploy import render_restfulapi, render_dashboard, render_storagemanager, render_repairmanager
 from cloud_init_deploy import check_buildable_images, push_docker_images
 
-FILE_MAP_PATH = 'deploy/cloud-config/file_map.yaml'
+FILE_MAP_PATH = "deploy/cloud-config/file_map.yaml"
+ENV_CNF_YAML = "config.yaml"
```
Anbang-Hu, line 28: These constants appear many times. It's better to put them into `constants.py`.

PR #1013, opened 2020-04-08
Rotate a new log blob if current one is full

Gerhut:


src/docker-images/azure-blob-adapter/wsgi.py
```diff
@@ -30,16 +30,21 @@ def application(request):
...
-            try:
-                blob = request.get_data()
-                count = request.content_length
+            blob = request.get_data()
+            count = request.content_length
+
+            suffix = 0
+            blob_name = tag
+
+            for _ in range(10):
```
Anbang-Hu, line 47: What happens if a job has logs exceeding 10 blobs?

Anbang-Hu: We can use `list_blobs` or `list_blob_names` to find the latest blob to append to so that we don't go over a limited number of blobs each time. An example:
```
>>> list(append_blob_service.list_blob_names(container_name="test"))
['job1/job1', 'job1/job1.1', 'job1/job1.2', 'job1/job1.3', 'job1/job1.4', 'job1/job1.5', 'job1/job1.6', 'job1/job1.7', 'job1/job1.8', 'job1/job1.9', 'job2/job2', 'job2/job2.1', 'job2/job2.2', 'job2/job2.3']
>>> list(append_blob_service.list_blob_names(container_name="test", prefix="job2/"))
['job2/job2', 'job2/job2.1', 'job2/job2.2', 'job2/job2.3']
>>> list(append_blob_service.list_blob_names(container_name="test", prefix="job3/"))
[]
>>>
```

`list_blobs` will return blobs whose `properties` contains `last_modified`

PR #1022, opened 2020-04-14
Cluster GPU accounting should include pods without node assignment

Anbang-Hu:
- [x] Add this logic to watchdog

### DIFF
--- a/src/docker-images/watchdog/src/watchdog.py
+++ b/src/docker-images/watchdog/src/watchdog.py
@@ -767,6 +767,20 @@ def process_nodes(k8s_api_addr, ca_path, headers, pods_info, cluster_gpu_info):
     return process_nodes_status(nodes_object, pods_info, cluster_gpu_info)
 
 
+def process_unscheduled_pods(pods_info, cluster_gpu_info):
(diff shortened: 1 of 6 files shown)


src/docker-images/watchdog/src/watchdog.py
```diff
@@ -767,6 +767,20 @@ def process_nodes(k8s_api_addr, ca_path, headers, pods_info, cluster_gpu_info):
...
+    preemptable_used_gpu = 0
+    unscheduled_pods = pods_info.get("unscheduled", [])
+    for pod in unscheduled_pods:
+        if pod.preemptable:
+            preemptable_used_gpu += pod.gpu
+        else:
+            used_gpu += pod.gpu
+
+    cluster_gpu_info.available -= used_gpu
+    cluster_gpu_info.preemptable_available -= (used_gpu + preemptable_used_gpu)
```
Anbang-Hu, line 781: @xudifsd can you help check if this logic is correct?

xudifsd: seems correct

PR #1024, opened 2020-04-14
Maintaindb

YinYangOfDao:
support more convenient database interaction,
might need to add columns_to_ignore to dumped yaml before pushing.
e.g.
```
./ctl.py db connect
./ctl.py db pull vc
```
(edit vc.yaml)
```
./ctl.py db push vc
```

TODO: this is 1st version, use in backend.
It's also necessary to have a restfulapi version (admin tool)

### DIFF
--- a/src/ClusterBootstrap/ctl.py
+++ b/src/ClusterBootstrap/ctl.py
@@ -258,42 +258,62 @@ def maintain_db(config, args):
     """
     push/pull a table to/from DB
     """
-    assert args.nargs[0] in ["pull", "push", "connect"], "invalid action."
+    subcommand = args.nargs[0]
+    assert subcommand in ["pull", "push", "connect"], "invalid action."
     host = config["mysql_node"]
     user = config["mysql_username"]
     password = config["mysql_password"]
-    if args.nargs[0] == "connect":
+    if subcommand == "connect":
         os.system("mysql -h {} -u {} -p{}".format(host, user, password))
     else:
         database = "DLWSCluster-{}".format(config["clusterId"])
-        assert args.nargs[1] in ["vc", "acl"], "invalid table."
+        table_name = args.nargs[1]
+        assert table_name in ["vc", "acl"], "invalid table."
         
         if args.verbose:
             print("connecting to {}@{}, DB {}".format(user, host, database))
         conn = connector.connect(user=user, password=password,
                host=host, database=database)
-        if args.nargs[0] == "pull":
-            sql = "SELECT * from {}".format(args.nargs[1])
-        cursor = conn.cursor()
-        cursor.execute(sql)
-        col_names = [col[0] for col in cursor.description]
-        serialized_rows = []
-        rows = cursor.fetchall()
+        if subcommand == "pull":
+            sql = "SELECT * from {}".format(table_name)
+            cursor = conn.cursor()
+            cursor.execute(sql)
+            col_names = [col[0] for col in cursor.description]
+            serialized_rows = []
+            rows = cursor.fetchall()
+            for row in rows:
+                serialized_row = {}
+                for i, v in enumerate(row):
+                    try:
+                        serialized_row[col_names[i]] = json.loads(v)
+                    # JSONDecodeError
+                    except:
+                        serialized_row[col_names[i]] = v
+                serialized_rows.append(serialized_row)
+            table_config = {"col_names": col_names, "rows": serialized_rows}
+            dump2 = args.output if args.output else "{}.yaml".format(table_name)
+            with open(dump2, "w") as wf:
+                yaml.safe_dump(table_config, wf)
+        elif subcommand == "push":
+            sql = "DELETE from {}".format(table_name)
+            cursor = conn.cursor()
+            cursor.execute(sql)
+            with open("{}.yaml".format(table_name)) as f:
+                table_config = yaml.safe_load(f)
+            col_names = table_config["col_names"]
+            cols_2_ignore = table_config.get("columns_to_ignore", ["time"])
+            cols_filtered = [col for col in col_names if col not in cols_2_ignore]
+            cols_str = ", ".join(cols_filtered)
+            for row in table_config["rows"]:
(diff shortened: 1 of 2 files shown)


src/ClusterBootstrap/deploy.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
-rm -rf deploy/!\(bin\) cloudinit* !\(config\).yaml
+shopt -s extglob
```
Anbang-Hu, line 2: What is this for?

src/ClusterBootstrap/ctl.py
```diff
@@ -253,6 +254,67 @@ def render_template_or_dir(config, args):
...
+    assert subcommand in ["pull", "push", "connect"], "invalid action."
+    host = config["mysql_node"]
+    user = config["mysql_username"]
+    password = config["mysql_password"]
+    if subcommand == "connect":
+        os.system("mysql -h {} -u {} -p{}".format(host, user, password))
+    else:
+        database = "DLWSCluster-{}".format(config["clusterId"])
+        table_name = args.nargs[1]
+        assert table_name in ["vc", "acl"], "invalid table."
```
Anbang-Hu, line 271: We should list what tables are valid in this operation

src/ClusterBootstrap/ctl.py
```diff
@@ -253,6 +254,67 @@ def render_template_or_dir(config, args):
         utils.render_template(src, dst, config)
 
 
+def maintain_db(config, args):
```
Anbang-Hu, line 257: Can you put this function into a separate common file so that `deploy.py` can use it? Can you also make it work in `deploy.py`?

src/ClusterBootstrap/ctl.py
```diff
@@ -253,6 +254,67 @@ def render_template_or_dir(config, args):
...
+                serialized_row = {}
+                for i, v in enumerate(row):
+                    try:
+                        serialized_row[col_names[i]] = json.loads(v)
+                    # JSONDecodeError
+                    except:
+                        serialized_row[col_names[i]] = v
+                serialized_rows.append(serialized_row)
+            table_config = {"col_names": col_names, "rows": serialized_rows}
+            dump2 = args.output if args.output else "{}.yaml".format(table_name)
```
Anbang-Hu, line 294: Can we use a clearer name like `output_file`?

src/ClusterBootstrap/ctl.py
```diff
@@ -253,6 +254,67 @@ def render_template_or_dir(config, args):
...
+            with open(dump2, "w") as wf:
+                yaml.safe_dump(table_config, wf)
+        elif subcommand == "push":
+            sql = "DELETE from {}".format(table_name)
+            cursor = conn.cursor()
+            cursor.execute(sql)
+            with open("{}.yaml".format(table_name)) as f:
```
Anbang-Hu, line 301: Symmetrically, I think we should handle user specified `args.in`.

src/ClusterBootstrap/ctl.py
```diff
@@ -253,6 +254,67 @@ def render_template_or_dir(config, args):
...
+                table_config = yaml.safe_load(f)
+            col_names = table_config["col_names"]
+            cols_2_ignore = table_config.get("columns_to_ignore", ["time"])
```
Anbang-Hu, line 304: Will the time be dumped into the file when `db pull`?

src/ClusterBootstrap/ctl.py
```diff
@@ -253,6 +254,67 @@ def render_template_or_dir(config, args):
...
+            cols_filtered = [col for col in col_names if col not in cols_2_ignore]
+            cols_str = ", ".join(cols_filtered)
+            for row in table_config["rows"]:
+                vals = ", ".join(["'{}'".format(json.dumps(row[col])) for col in cols_filtered])
+                sql = "INSERT INTO `{}` ({}) VALUES ({})".format(table_name, cols_str, vals)
```
Anbang-Hu, line 309: I don't think this will work if I need to update the entry.

src/ClusterBootstrap/deploy.sh
YinYangOfDao: clean previous deployment, keep large-size universal binaries and config.yaml only.

src/ClusterBootstrap/ctl.py
YinYangOfDao: ok, which file do you prefer? ClusterBootstrap/utils.py?

src/ClusterBootstrap/ctl.py
YinYangOfDao: is there a list about what tables should be valid? are there any tables no longer in use.?

src/ClusterBootstrap/ctl.py
YinYangOfDao: yes. but when we update, we should use the DB generated real-time, so I discard this column

src/ClusterBootstrap/ctl.py
YinYangOfDao: actually I thought about this and would rather keep it in ctl.py, instead of utils.py, as it reads deployment config. if it reads in deployment config, ctl.py should be the file to keep this, and utils.py is mainly for those that is not tangled with configs.
also for deploy.py, since config files are not compatible, it would be faster to convert old config files to fit new pipeline schema

src/ClusterBootstrap/ctl.py
YinYangOfDao: sure, updated.

src/ClusterBootstrap/ctl.py
YinYangOfDao: updated

src/ClusterBootstrap/ctl.py
YinYangOfDao: delete all rows and then add based on yaml file, to better support db config editing. if you prefer, we can also set auto_increment variable ID to 1 after deleting rows.

PR #1027, opened 2020-04-15
[RepairManager] add etcd sidecar container to repairmanager

debbie-alaine:
Adding a sidecar docker container to repairmanager pod to serve as standalone etcd server. This will be accessed by repairmanager agent on worker nodes vi the infra node private IP.

Highlights:
-- configurations for etcd server are in rendered etcd.conf.yaml, which should be copied to persistent storage (/etc/RepairManager/config)
-- snapshots are also saved in persistent storage
-- uses apt-get to install etcd-server on ubuntu base image

Eventually, we would want to add authentication to access the etcd server as it is accessible from anywhere in cluster via the configured port. @hongzhili mentioned we could add this later.

@YinYangOfDao @Anbang-Hu @xudifsd


src/ClusterBootstrap/template/RepairManager/etcd.conf.yaml
```diff
@@ -0,0 +1,21 @@
+# This is the configuration file for the etcd server
+
+# Path to the data directory
+data-dir: {{cnf['repair-manager']['etcd']['data-dir']}}
```
Anbang-Hu, line 4: Can you add a default value for `"data-dir"` in `params.py`? line 60.

PR #1028, opened 2020-04-15
support auto-scale for inference job

leigaoms:
1. Provide RestAPI interface;
2. JobManager deployment could auto-scale; framework launcher does not support yet

### DIFF
--- a/src/ClusterManager/job_launcher.py
+++ b/src/ClusterManager/job_launcher.py
@@ -252,6 +252,26 @@ class Launcher(object):
         return api_response
 
     @record
+    def _get_deployment(self, name):
+        api_response = self.k8s_AppsAPI.read_namespaced_deployment_scale(
+            namespace=self.namespace,
+            pretty=self.pretty,
+            name=name
+        )
+        logger.debug("Get pods: {}".format(api_response))
+        return api_response
+
+    @record
+    def _patch_deployment(self, name, body):
+        api_response = self.k8s_AppsAPI.patch_namespaced_deployment_scale(
+            namespace=self.namespace,
+            pretty=self.pretty,
+            name=name,
+            body=body
+        )
+        return api_response
+
+    @record
     def _create_service(self, body):
         api_response = self.k8s_CoreAPI.create_namespaced_service(
             namespace=self.namespace,
@@ -786,6 +806,9 @@ class LauncherStub(Launcher):
             logger.error("Kill job failed with errors: {}".format(errors))
             return False
 
+    def scale_job(self, job):
+        pass
+
 
 class PythonLauncher(Launcher):
     def __init__(self, pool_size=3):
@@ -864,7 +887,7 @@ class PythonLauncher(Launcher):
             for pod in pods
             if pod["kind"] == "Deployment"
         ]
-        self._cleanup_deployment(pod_names)
+        self._cleanup_deployment(deployment_names)
         created = []
         for pod in pods:
             if pod["kind"] == "Pod":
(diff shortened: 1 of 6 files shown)


src/RestAPI/dlwsrestapi.py
```diff
@@ -309,6 +309,26 @@ def get(self):
         return generate_response(ret)
 
 
+@api.resource("/ScaleInferenceJob")
```
xudifsd, line 312: Maybe use `/ScaleJob`, do not be so specific about which job can be scaled in API

src/ClusterManager/job_launcher.py
```diff
@@ -786,6 +806,9 @@ def kill_job(self, job_id, desired_state="killed"):
             logger.error("Kill job failed with errors: {}".format(errors))
             return False
 
+    def scale_job(self, job):
+        pass
```
xudifsd, line 810: Maybe add a logger.warning to show controller doesn't support it now.

xudifsd:
Hope you can add a test case [here](https://github.com/microsoft/DLWorkspace/blob/dltsdev/src/ClusterManager/test/test_inference_job.py) to test this functionality. You may want to add branch like [this](https://github.com/microsoft/DLWorkspace/blob/dltsdev/src/ClusterManager/test/test_regular_job.py#L649) to skip the test case if it's `controller`.

It can be run with parameters `./main.py --rest http://dltshub-int.redmond.corp.microsoft.com:5001 --vc platform --email xxx@microsoft.com --uid 1234 --config ~/dev/Deployment/Azure-EastUS-P40-Dev1/` please replace email and uid and config parameter when you run. With your email, uid get from identity table and config to backuped cluster config.

I want to make sure all test cases not broken whenever new feature added.

src/RestAPI/dlwsrestapi.py
```diff
@@ -309,6 +309,26 @@ def get(self):
...
+        self.get_parser.add_argument("jobId", required=True)
+        self.get_parser.add_argument("userName", required=True)
+        self.get_parser.add_argument("resourcegpu", required=True)
+
+    def get(self):
+        args = self.get_parser.parse_args()
+        job_id = args["jobId"]
+        username = args["userName"]
+        resourcegpu = int(args["resourcegpu"])
+        msg, statusCode = JobRestAPIUtils.scale_inference_job(username, job_id, resourcegpu)
```
Anbang-Hu, line 325: Let's use snake case for variable: `status_code`

src/utils/JobRestAPIUtils.py
```diff
@@ -481,6 +481,40 @@ def resume_job(username, job_id):
...
+            msg = "Only inference job could be scaled, current job %s is %s" % (job_id, job_type)
+            logger.error(msg)
+            return msg, 403
+
+        job_params["resourcegpu"] = resourcegpu
+        dataHandler.UpdateJobTextFields({"jobId": job_id},
+                                        {"jobParams": base64encode(json.dumps(job_params))})
+        return "Success", 200
+    except Exception as e:
+        logger.error("Scale inference job exception, ex: %s", str(e))
```
Anbang-Hu, line 514: `logger.exception` will log exception. You don't need to specify `str(e)`.

src/ClusterManager/job_launcher.py
leigaoms: This will dump too many logs, just skip it now.

src/ClusterManager/test/test_inference_job.py
```diff
@@ -70,3 +70,38 @@ def test_inference_job_running(args):
...
+        deployment_name = job_id + "-deployment"
+        deployment = utils.kube_get_deployment(args.config, "default", deployment_name)
+        assert 1 == deployment.spec.replicas
+
+        desired_replicas = 2
+        logger.info("scale up job %s to %d" % (job_id, desired_replicas))
+        resp = utils.scale_job(args.rest, args.email, job_id, desired_replicas)
+        assert "Success" == resp
+
+        time.sleep(30)
```
xudifsd, line 96: I would prefer block_until_replica_count_is with a timeout parameter in test case. This will wait at least 30s.

src/ClusterManager/test/test_inference_job.py
```diff
@@ -70,3 +70,38 @@ def test_inference_job_running(args):
...
+                                                args.uid, args.vc, cmd="sleep 600")
+
+    with utils.run_job(args.rest, job_spec) as job:
+        job_id = job.jid
+        state = job.block_until_state_not_in(
+            {"unapproved", "queued", "scheduling"})
+        assert state == "running"
+
+        deployment_name = job_id + "-deployment"
+        deployment = utils.kube_get_deployment(args.config, "default", deployment_name)
```
xudifsd, line 88: I think we can relax assumption about it should be a deployment here. We can count pods with label jobId=xxx here to check if it scaled up or down. Since controller doesn't create deployment.

xudifsd:
Anyway, it's good to have a test case. I will modify this test case when I have time. You can merge this for now.

src/ClusterManager/test/test_inference_job.py
leigaoms: The problem is if cluster does not have enough resource, the number of pod may never reach to desired_replicas. We can just make sure deployment.replica is correctly set, but the real replica also impacted by cluster status.

PR #1032, opened 2020-04-15
do not connect to k8s api server when it's regular job

xudifsd:
reduce dependency on api server


src/ClusterManager/test/utils.py
```diff
@@ -64,6 +69,49 @@ def wrapped(*args, **kwargs):
...
+
+        for p in pool:
+            p.join()
+    except KeyboardInterrupt:
+        pass
+
+    def getter(queue):
+        if queue.empty():
+            return None
+        return queue.get(False)
```
Anbang-Hu, line 110: `queue.get_nowait()` is clearer?

PR #1036, opened 2020-04-17
add cordon/uncordon

YinYangOfDao:
add cordon/uncordon, to avoid uncordoning nodes cordoned manually by other devs
when uncordon nodes, use ./ctl.py uncordon or ./deploy.py uncordon:
```
./deploy.py cordon <node name> "NFS test"
```
and you will see output:
```
node/<node name> annotated
node/<node name> cordoned
```
if you want to uncordon a node: 
```
./deploy.py uncordon <node name>
```
or 
```
./ctl.py cordon <node name> "NFS test"
./ctl.py uncordon <node name>
```
if it is already cordoned with this tool, you will be prompted.

### DIFF
--- a/src/ClusterBootstrap/ctl.py
+++ b/src/ClusterBootstrap/ctl.py
@@ -5,8 +5,10 @@ import re
 import sys
 import time
 import json
+import pytz
 import utils
 import yaml
+import datetime
 import argparse
 import textwrap
 import random
@@ -16,6 +18,7 @@ os.chdir(cwd)
 
 sys.path.append("../utils")
 
+from pathlib import Path
 from ConfigUtils import *
 from constants import FILE_MAP_PATH, ENV_CNF_YAML, STATUS_YAML
 
@@ -316,6 +319,40 @@ def maintain_db(config, args):
         cursor.close()
 
 
+def cordon(config, args):
+    home_dir = str(Path.home())
+    DLTS_admin_config_path = os.path.join(home_dir, ".dlts-admin.yaml")
+    DLTS_admin_config_path = config.get(
+        "DLTS_admin_config_path", DLTS_admin_config_path)
+    if os.path.exists(DLTS_admin_config_path):
+        with open(DLTS_admin_config_path) as f:
+            admin_name = yaml.safe_load(f)["admin_name"]
+    else:
+        admin_name = args.admin
+        assert admin_name is not None and admin_name, "specify admin name by"\
+        "--admin or in ~/.dlts-admin.yaml"
+    now = datetime.datetime.now(pytz.timezone("UTC"))
+    timestr = now.strftime("%Y/%m/%d %H:%M:%S %Z")
+    node = args.nargs[0]
+    note = " ".join(args.nargs[1:])
+    annotation = "cordoned by {} at {}, {}".format(admin_name, timestr, note)
+    k8s_cmd = "annotate node {} --overwrite cordon-note='{}'".format(
+        node, annotation)
+    run_kubectl(config, args, [k8s_cmd])
+    run_kubectl(config, args, ["cordon {}".format(node)])
+
+
+def uncordon(config, args):
+    node = args.nargs[0]
+    query_cmd = "get nodes {} -o=jsonpath=\'{{.metadata.annotations.cordon-note}}\'".format(node)
+    output = run_kubectl(config, args, [query_cmd], need_output=True)
+    if output and not args.force:
+        print("node annotated, if you are sure that you want to uncordon it, "\
+            "please specify --force or use ./ctl.py kubectl cordon <node> to cordon")
+    else:
+        run_kubectl(config, args, ["uncordon {}".format(node)])
+
+
 def run_command(args, command):
     config = load_config_4_ctl(args, command)
     if command == "restorefromdir":
@@ -359,6 +396,10 @@ def run_command(args, command):
             push_docker_images(args, config)
     elif command == "db":
         maintain_db(config, args)
+    elif command == "cordon":
+        cordon(config, args)
+    elif command == "uncordon":
+        uncordon(config, args)
     else:
         print("invalid command, please read the doc")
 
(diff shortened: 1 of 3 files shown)


src/ClusterBootstrap/ctl.py
```diff
@@ -316,6 +319,40 @@ def maintain_db(config, args):
         cursor.close()
 
 
+def cordon(config, args):
+    home_dir = str(Path.home())
+    DLTS_admin_config_path = os.path.join(home_dir, ".dlts-admin.yaml")
```
Anbang-Hu, line 324: `dlts_admin_config_path`

src/ClusterBootstrap/ctl.py
```diff
@@ -316,6 +319,40 @@ def maintain_db(config, args):
         cursor.close()
 
 
+def cordon(config, args):
```
Anbang-Hu, line 322: Can these functions be reused by`ctl.py`?

src/ClusterBootstrap/ctl.py
```diff
@@ -316,6 +319,40 @@ def maintain_db(config, args):
...
+    DLTS_admin_config_path = os.path.join(home_dir, ".dlts-admin.yaml")
+    DLTS_admin_config_path = config.get(
+        "DLTS_admin_config_path", DLTS_admin_config_path)
+    if os.path.exists(DLTS_admin_config_path):
+        with open(DLTS_admin_config_path) as f:
+            admin_name = yaml.safe_load(f)["admin_name"]
+    else:
+        admin_name = args.admin
+        assert admin_name is not None and admin_name, "specify admin name by"\
+        "--admin or in ~/.dlts-admin.yaml"
```
Anbang-Hu, line 333: `specify admin_name by ...`

src/ClusterBootstrap/ctl.py
```diff
@@ -316,6 +319,40 @@ def maintain_db(config, args):
...
+    run_kubectl(config, args, ["cordon {}".format(node)])
+
+
+def uncordon(config, args):
+    node = args.nargs[0]
+    query_cmd = "get nodes {} -o=jsonpath=\'{{.metadata.annotations.cordon-note}}\'".format(node)
+    output = run_kubectl(config, args, [query_cmd], need_output=True)
+    if output and not args.force:
+        print("node annotated, if you are sure that you want to uncordon it, "\
+            "please specify --force or use ./ctl.py kubectl cordon <node> to cordon")
```
Anbang-Hu, line 351: `./{}`.format(__file__)

src/ClusterBootstrap/ctl.py
```diff
@@ -395,10 +436,14 @@ def run_command(args, command):
...
                         help="Build docker without cache",
                         action="store_true")
+    parser.add_argument("--admin",
+                        help="Name of admin that execute this script")
     parser.add_argument("command",
                         help="See above for the list of valid command")
     parser.add_argument('nargs', nargs=argparse.REMAINDER,
                         help="Additional command argument")
+    parser.add_argument(
+        '-d', '--dryrun', help='Dry run -- no actual execution', action="store_true")
```
Anbang-Hu, line 446: This parameter doesn't seem used?

src/ClusterBootstrap/deploy.py
```diff
@@ -2470,6 +2484,39 @@ def kubernetes_get_node_name(node):
         kube_node_name = node
     return kube_node_name
 
+def cordon(config, args):
```
Anbang-Hu, line 2487: Duplication of the ones in `ctl.py`?

src/ClusterBootstrap/deploy.py
```diff
@@ -2452,11 +2457,20 @@ def run_kube(prog, commands):
                                                                                                                              config["k8sAPIport"], "./deploy/ssl/ca/ca.pem", "./deploy/ssl/kubelet/apiserver-key.pem", "./deploy/ssl/kubelet/apiserver.pem", one_command))
     if verbose:
         print(kube_command)
-    os.system(kube_command)
+    if need_output:
+        # we may want to dump command to another file instead of args.output, when we don't want to mix k8s commands with others
+        print("output specified")
+        output = utils.execute_or_dump_locally(kube_command, verbose, False, '')
+        if not args.verbose:
```
Anbang-Hu, line 2464: `args.verbose` -> `verbose`? Why do we want to print it when not verbose? We should print it when verbose as well right?

src/ClusterBootstrap/deploy.py
```diff
@@ -2452,11 +2457,20 @@ def run_kube(prog, commands):
                                                                                                                              config["k8sAPIport"], "./deploy/ssl/ca/ca.pem", "./deploy/ssl/kubelet/apiserver-key.pem", "./deploy/ssl/kubelet/apiserver.pem", one_command))
     if verbose:
         print(kube_command)
-    os.system(kube_command)
+    if need_output:
+        # we may want to dump command to another file instead of args.output, when we don't want to mix k8s commands with others
+        print("output specified")
```
Anbang-Hu, line 2462: Why do we need this print?

src/ClusterBootstrap/deploy.py
YinYangOfDao: when verbose, it would be printed at line 2463.
if not verbose, this is for kubectl . remember that we decide to output anyway when it's kubectl? so devs don't need to specify -v

src/ClusterBootstrap/deploy.py
YinYangOfDao: actually not. the old pipeline didn't take config/args as input variables, and used global variables, which is a very important factor that lead to refactoring. 
difference lies where run_kubectl is invoked, in old pipeline, config/args are not passed to run_kubectl.

src/ClusterBootstrap/ctl.py
YinYangOfDao: you mean cordon/uncordon? yes, as they are defined in this file. Currently this is the only place that they are invoked. if you mean reuse in deploy.py, then no. (check my reply below)

src/ClusterBootstrap/ctl.py
YinYangOfDao: it's used. previously we don't have this parameter in ctl.py because we basically only set need_output in other files, not ctl.py. Now we need output to filter annotated nodes, so this parameter is used.

src/ClusterBootstrap/ctl.py
YinYangOfDao: corrected

src/ClusterBootstrap/ctl.py
YinYangOfDao: corrected

src/ClusterBootstrap/deploy.py
YinYangOfDao: deleted

src/ClusterBootstrap/ctl.py
YinYangOfDao: corrected

PR #1039, opened 2020-04-18
[RepairManager] add RepairManagerAgent service

debbie-alaine:
The repair-manager agent runs on each worker node, polling etcd for signal to reboot.

When reboot signal is found, repairmanager agent deletes the key and reboots the node. The repairmanager agent pod will return when the node is back up and running (This takes a few minutes). There is also dry-run capability.

@hongzhili @Anbang-Hu @YinYangOfDao @xudifsd 

MicrosoftTeams-image

### DIFF
--- a/src/ClusterBootstrap/params.py
+++ b/src/ClusterBootstrap/params.py
@@ -62,7 +62,9 @@ default_config_parameters = {
         "prometheus-ip": "localhost",
         "prometheus-port": 9091,
         "etcd": {
-            "data-dir": "/etc/RepairManager/etcd"
+            "data-dir": "/etc/RepairManager/etcd",
+            "peer-port": 2381,
+            "client-port": 2382
         }
     },
 
@@ -91,6 +93,7 @@ default_config_parameters = {
     "storagemanager": "storagemanager",
     "repairmanager": "repairmanager",
     "repairmanageretcd": "repairmanageretcd",
+    "repairmanageragent": "repairmanageragent",
     "ssh_cert": "./deploy/sshkey/id_rsa",
     "admin_username": "core",
     # the path of where dfs/nfs is source linked and consumed on each node,
@@ -697,7 +700,7 @@ default_config_parameters = {
         # There is no udp port requirement for now
         #"udp_port_ranges": "25826",
         "inter_connect": {
-            "tcp_port_ranges": "22 1443 2379 3306 5000 8086 9092 9114 9200 9300 10250",
+            "tcp_port_ranges": "22 1443 2379 2382 3306 5000 8086 9092 9114 9200 9300 10250",
             # Need to white list dev machines to connect
             # "source_addresses_prefixes": [ "52.151.0.0/16"]
         },
--- a/src/ClusterBootstrap/params.py
+++ b/src/ClusterBootstrap/params.py
@@ -293,6 +293,7 @@ default_config_parameters = {
         'worker': {
             "worker": "active",
             "beta.kubernetes.io/os": "linux",
+            "repairmanageragent": "active"
             },
         'mysqlserver': {
             "mysql-server": "active"
--- a/src/ClusterBootstrap/services/repairmanageragent/repairmanageragent.yaml
+++ b/src/ClusterBootstrap/services/repairmanageragent/repairmanageragent.yaml
@@ -1,0 +1,90 @@
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: repairmanageragent
+  namespace: default
+spec:
+  selector:
+    matchLabels:
+      repairmanageragent-node: pod
+  template:
+    metadata:
+      name: repairmanageragent
+      labels:
+        repairmanageragent-node: pod
+        app: repairmanageragent
+    spec:
+      {% if cnf["dnsPolicy"] %}
+      dnsPolicy: {{cnf["dnsPolicy"]}}
+      {% endif %}
+      hostPID: true
+      nodeSelector:
+        repairmanageragent: active
+      containers:
+      - name: repairmanageragent
+        image: {{cnf["worker-dockerregistry"]}}/{{cnf["dockerprefix"]}}/{{cnf["repairmanageragent"]}}:{{cnf["dockertag"]}}
+        securityContext:
+          capabilities:
+            add: ["SYS_BOOT"]
+        command: ["/run.sh"]
+        imagePullPolicy: Always
+        env:
+        - name: NODE_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        - name: ETCD_SERVER
+          value: "{{cnf['kubernetes_master_node'][0]}}"
+        - name: ETCD_PORT
+          value: "{{cnf['repair-manager']['etcd']['client-port']}}"
+        volumeMounts:
+        - mountPath: /etc/kubernetes/ssl
+          name: certs
+        - mountPath: /etc/kubernetes/restapi-kubeconfig.yaml
+          name: kubeconfig
+        - mountPath: {{cnf["storage-mount-path"]}}/work
+          name: dlwsdatawork
+        - mountPath: {{cnf["storage-mount-path"]}}/storage
+          name: dlwsdatadata
+        - mountPath: {{cnf["storage-mount-path"]}}/jobfiles
+          name: dlwsdatajobfiles
+        - mountPath: {{cnf["dltsdata-storage-mount-path"]}}
+          name: dltsdata
+        - mountPath: {{cnf["folder_auto_share"]}}
+          name: folderautoshare
+        - mountPath: /var/log/dlworkspace
+          name: log
+      {% if cnf["private_docker_registry_username"] %}
+      imagePullSecrets:
+      - name: svccred
+      {% endif %}
+      volumes:
+      - name: certs
+        hostPath:
+          path: /etc/kubernetes/ssl
+      - name: kubeconfig
+        hostPath:
+          path: /etc/kubernetes/restapi-kubeconfig.yaml
+      - name: dlwsdatawork
+        hostPath:
+          path: {{cnf["storage-mount-path"]}}/work
+      - name: dlwsdatadata
+        hostPath:
+          path: {{cnf["storage-mount-path"]}}/storage
+      - name: dlwsdatajobfiles
+        hostPath:
+          path: {{cnf["storage-mount-path"]}}/jobfiles
+      - name: dltsdata
+        hostPath:
+          path: {{cnf["dltsdata-storage-mount-path"]}}
+      - name: folderautoshare
+        hostPath:
+          path: {{cnf["folder_auto_share"]}}
+      - name: log
+        hostPath:
+          path: /var/log/clustermanager
+      tolerations:
+      - key: CriticalAddonsOnly
+        operator: Exists
+      - key: node-role.kubernetes.io/master
+        effect: NoSchedule
--- a/src/ClusterBootstrap/services/repairmanageragent/repairmanageragent.yaml
+++ b/src/ClusterBootstrap/services/repairmanageragent/repairmanageragent.yaml
@@ -82,7 +82,7 @@ spec:
           path: {{cnf["folder_auto_share"]}}
       - name: log
         hostPath:
-          path: /var/log/clustermanager
+          path: /var/log/dlworkspace
       tolerations:
       - key: CriticalAddonsOnly
         operator: Exists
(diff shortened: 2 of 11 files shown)


src/ClusterBootstrap/services/repairmanageragent/repairmanageragent.yaml
```diff
@@ -0,0 +1,90 @@
...
+        imagePullPolicy: Always
+        env:
+        - name: NODE_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        - name: ETCD_SERVER
+          value: "{{cnf['kubernetes_master_node'][0]}}"
+        - name: ETCD_PORT
+          value: "{{cnf['repair-manager']['etcd']['client-port']}}"
```
debbie-alaine, line 39: environment variables passed to the docker container allow the repairmanager agent to get and update key values in etcd server.

src/ClusterBootstrap/services/repairmanageragent/repairmanageragent.yaml
```diff
@@ -0,0 +1,90 @@
...
+      {% endif %}
+      hostPID: true
+      nodeSelector:
+        repairmanageragent: active
+      containers:
+      - name: repairmanageragent
+        image: {{cnf["worker-dockerregistry"]}}/{{cnf["dockerprefix"]}}/{{cnf["repairmanageragent"]}}:{{cnf["dockertag"]}}
+        securityContext:
+          capabilities:
+            add: ["SYS_BOOT"]
```
debbie-alaine, line 28: Added SYS_BOOT linux capability to allow for reboot syscall from docker container

src/RepairManagerAgent/main.py
```diff
@@ -0,0 +1,47 @@
...
+
+with open('./logging.yaml', 'r') as log_file:
+    log_config = yaml.safe_load(log_file)
+
+logging.config.dictConfig(log_config)
+logger = logging.getLogger(__name__)
+
+def reboot_node():
+    os.system('sync')
+    os.system('reboot -f')
```
debbie-alaine, line 16: method to reboot host node

src/RepairManagerAgent/main.py
```diff
@@ -0,0 +1,47 @@
...
+
+def Run():
+    while True:
+        etcd_server = os.environ["ETCD_SERVER"]
+        etcd_port = os.environ["ETCD_PORT"]
+        node_name = os.environ["NODE_NAME"]
+        etcd_url = f'http://{etcd_server}:{etcd_port}/v2/keys/{node_name}/reboot'
```
debbie-alaine, line 23: the key is {NODE_NAME}/reboot, so each pod on each worker node will be polling for a different key

src/RepairManagerAgent/main.py
```diff
@@ -0,0 +1,47 @@
...
+            response = requests.get(etcd_url, timeout=5)
+            r_json = response.json()
+
+            if r_json is not None:
+                logger.debug(r_json)
+
+                if 'node' in r_json and 'key' in r_json['node']:
+                    value = r_json['node']['value']
+                    if value == 'True':
+                        requests.put(f'{etcd_url}', data={'value':'False'})
```
debbie-alaine, line 36: set the reboot key back to False before rebooting so that the node doesn't get into a loop of rebooting... could also do this from the repair manager side...

src/ClusterBootstrap/params.py
```diff
@@ -91,6 +93,7 @@
     "storagemanager": "storagemanager",
     "repairmanager": "repairmanager",
     "repairmanageretcd": "repairmanageretcd",
+    "repairmanageragent": "repairmanageragent",
```
YinYangOfDao, line 96: where is this value rendered?

YinYangOfDao: Also in params.py, default_kube_labels_by_node_role["worker"] need to have "repairmanageragent": "active"

debbie-alaine: It is rendered in the dockerfile: https://github.com/microsoft/DLWorkspace/blob/775b8f61e74ffb240dd12b0d80616bfb1ba90f4a/src/ClusterBootstrap/services/repairmanageragent/repairmanageragent.yaml#L25

YinYangOfDao: @Anbang-Hu , I noticed that pod yaml under services have different way to specify the image, e.g., storagemanager & restfulapi use {{cnf[]}} and have default value in params.py, while watchdog hard-coded the image url. I suggest later we refactor and formalize this.

debbie-alaine: @YinYangOfDao just FYI I used jobmanager service as a reference for repairmanager service so this is why I have done this way. agree that there should be a formal way to do this for all services.

src/RepairManagerAgent/logging.yaml
```diff
@@ -0,0 +1,28 @@
...
+    class: logging.StreamHandler
+    level: DEBUG
+    formatter: simple
+    stream: ext://sys.stdout
+  file:
+    class : logging.handlers.RotatingFileHandler
+    formatter: simple
+    filename: repairmanageragent.log
+    # roll over at 10MB
+    maxBytes: 10240000
```
xudifsd, line 18: Make it 100M, repair manager agent log is quite important.

src/docker-images/RepairManagerAgent/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM ubuntu:18.04
```
xudifsd, line 1: Maybe you can use `python:3.7` base image, that will make repair manager agent image much smaller. For ref https://github.com/microsoft/DLWorkspace/blob/dltsdev/src/docker-images/watchdog/Dockerfile#L18

src/docker-images/RepairManagerAgent/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+RUN apt-get update && apt-get --no-install-recommends install -y \
+    python3.5 \
+    python3-pip \
+    systemd-sysv
+
+RUN pip3 install wheel
+RUN pip3 install setuptools
+RUN pip3 install requests
+RUN pip3 install pyyaml
```
xudifsd, line 12: Make it one line. Multiple lines will generate multiple image layers which will make pulling unncessary slow. For ref https://hackernoon.com/tips-to-reduce-docker-image-sizes-876095da3b34

src/RepairManagerAgent/main.py
```diff
@@ -0,0 +1,47 @@
...
+    os.system('sync')
+    os.system('reboot -f')
+
+def Run():
+    while True:
+        etcd_server = os.environ["ETCD_SERVER"]
+        etcd_port = os.environ["ETCD_PORT"]
+        node_name = os.environ["NODE_NAME"]
+        etcd_url = f'http://{etcd_server}:{etcd_port}/v2/keys/{node_name}/reboot'
+        logger.debug(etcd_url)
```
xudifsd, line 24: You may want to move environment get and this logging out from while. It will not change anyway and will output large chunk of logging.

src/RepairManagerAgent/logging.yaml
```diff
@@ -0,0 +1,28 @@
...
+handlers:
+  console:
+    class: logging.StreamHandler
+    level: DEBUG
+    formatter: simple
+    stream: ext://sys.stdout
+  file:
+    class : logging.handlers.RotatingFileHandler
+    formatter: simple
+    filename: repairmanageragent.log
```
xudifsd, line 16: `/var/log/dlworkspace/repairmanageragent.log`? I'm not sure where this logging file will be when you use relative path.

src/RepairManagerAgent/main.py
```diff
@@ -0,0 +1,47 @@
...
+
+            if r_json is not None:
+                logger.debug(r_json)
+
+                if 'node' in r_json and 'key' in r_json['node']:
+                    value = r_json['node']['value']
+                    if value == 'True':
+                        requests.put(f'{etcd_url}', data={'value':'False'})
+
+                        logger.debug("Attempting to reboot the node!")
```
xudifsd, line 38: Please use `info` or `warning`

src/RepairManagerAgent/logging.yaml
debbie-alaine: Ok, currently I have 10 backup files at ~10MB each. Should I instead have 1 file of 100MB?

src/RepairManagerAgent/logging.yaml
debbie-alaine: it gets logged inside the docker container... oops

src/RepairManagerAgent/logging.yaml
xudifsd: Please retain 10 backup count. 1G is not a big deal I think.

src/RepairManagerAgent/main.py
debbie-alaine: more on etcd data model: https://etcd.io/docs/v3.4.0/learning/data_model/

src/RepairManagerAgent/main.py
```diff
@@ -0,0 +1,47 @@
...
+
+    while True:
+        try: 
+            response = requests.get(etcd_url, timeout=5)
+            r_json = response.json()
+
+            if r_json is not None:
+                logger.debug(r_json)
+
+                if 'node' in r_json and 'key' in r_json['node']:
```
Anbang-Hu, line 33: `'key' in r_json['node']` or `'value' in r_json['node']`?

src/RepairManagerAgent/main.py
```diff
@@ -0,0 +1,47 @@
...
+                    value = r_json['node']['value']
+                    if value == 'True':
+                        requests.put(etcd_url, data={'value':'False'})
```
Anbang-Hu, line 36: When will this data be deleted in etcd?

src/ClusterBootstrap/params.py
```diff
@@ -62,7 +62,9 @@
         "prometheus-ip": "localhost",
         "prometheus-port": 9091,
         "etcd": {
-            "data-dir": "/etc/RepairManager/etcd"
+            "data-dir": "/etc/RepairManager/etcd",
+            "peer-port": 2381,
```
Anbang-Hu, line 66: Where is `peer-port` used?

src/RepairManagerAgent/main.py
debbie-alaine: this is checking if the key exists at all (the /{NODENAME}/reboot key)

debbie-alaine: This is the request response if key exists:
{'action': 'get', 'node': {'key': '/desandov-test-worker-cuvozl/reboot', 'value': 'False', 'modifiedIndex': 16, 'createdIndex': 16}}

This is the request response if key doesn't exist:
{"errorCode":100,"message":"Key not found","cause":"/desandov-test-worker-cuvozl/doesnotexist","index":18}

src/ClusterBootstrap/params.py
debbie-alaine: `peer-port` is used a couple times in the etcd.conf.yaml, which is used for starting up the etcd server.

https://github.com/microsoft/DLWorkspace/blob/e480f7f2f1868cb2ef5248363024d9ac14886e70/src/ClusterBootstrap/template/RepairManager/etcd.conf.yaml#L7

It's a required configuration to start up the etcd server, but it's not used otherwise (as there are currently no other etcd peers).

src/RepairManagerAgent/main.py
debbie-alaine: @Anbang-Hu, good point, should this key-value pair be completely deleted instead of overwritten? If so, maybe it would be best to delete it here.

src/ClusterBootstrap/services/repairmanageragent/repairmanageragent.yaml
```diff
@@ -0,0 +1,90 @@
...
+      - name: dltsdata
+        hostPath:
+          path: {{cnf["dltsdata-storage-mount-path"]}}
+      - name: folderautoshare
+        hostPath:
+          path: {{cnf["folder_auto_share"]}}
+      - name: log
+        hostPath:
+          path: /var/log/dlworkspace
+      tolerations:
```
YinYangOfDao, line 86: why do we use these tolerations?

debbie-alaine: actually not sure about this...

originally copied over from the jobmanager.yaml
https://github.com/microsoft/DLWorkspace/blob/05fff95805f574be28fe1dc42029a6e3b0befc72/src/ClusterBootstrap/services/jobmanager/jobmanager.yaml#L139

PR #1042, opened 2020-04-19
GC inactive jobs older than 180 days

Anbang-Hu:


xudifsd:
Have you tested this logic? Maybe change expiry day to 1 and run in dev1? It's pretty dangerous operation.

Anbang-Hu:
> Have you tested this logic? Maybe change expiry day to 1 and run in dev1? It's pretty dangerous operation.

@xudifsd This is tested on Dev1:
1. Inactive jobs with `lastUpdated` older than 180 days are successfully deleted.
2. Active jobs with `lastUpdated` older than 180 days (Manually updated in DB) are retained.

xudifsd:
Ok, cool

PR #1047, opened 2020-04-22
add nvsm health metric

xudifsd:
useful in DGX-2 machines


Anbang-Hu:
Any workaround using nvsm inside job-exporter docker?

xudifsd:
> Any workaround using nvsm inside job-exporter docker?

Didn't find any.

I tried to mount all required directories into container, but nvsm still have some hard coded path, I can not find a way around. I do read some nvidia documentation, trying to find an alternative.

PR #1048, opened 2020-04-22
Add assigned utils and aggregate.rules

Anbang-Hu:


src/docker-images/gpu-reporter/reporter.py
```diff
@@ -191,8 +192,11 @@ def get_monthly_idleness(prometheus_url):
 
                 if nonidle_time == 0:
                     user_val["nonidle_util"] = 0.0
+                    user_val["assigned_util"] = 0.0
```
xudifsd, line 195: This can be moved out of if branch. statement above is used to avoid divide by zero, but assigned_utils doesn't have such problem.

Anbang-Hu: Corrected.

PR #1050, opened 2020-04-23
Some admin utils for Azure clusters

Anbang-Hu:


xudifsd:
but, why the name `h`?

Anbang-Hu:
> but, why the name `h`?

health? I just want to type one character.

PR #1055, opened 2020-04-24
fix nvsm calling

xudifsd:
nvsm has very high latency. So adjust default timeout and buckets and decay_time.

### DIFF
--- a/src/docker-images/job-exporter/src/collector.py
+++ b/src/docker-images/job-exporter/src/collector.py
@@ -988,15 +988,15 @@ class DCGMCollector(Collector):
 class NVSMCollector(Collector):
     cmd_histogram = Histogram("cmd_nvsm_latency_seconds",
                               "Command call latency for nvsm (seconds)",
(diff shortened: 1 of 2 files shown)


src/docker-images/job-exporter/src/collector.py
```diff
@@ -988,15 +988,15 @@ def collect_impl(self):
...
+                              buckets=(32.0, 64.0, 128.0, 256.0, 512.0, 1024.0,
+                                       2048.0, 4096.0, float("inf")))
 
-    cmd_timeout = 300 # usually take around 200s
+    cmd_timeout = 1200
 
     def __init__(self, name, sleep_time, atomic_ref, iteration_counter):
         Collector.__init__(self, name, sleep_time, atomic_ref,
                            iteration_counter)
+        os.environ["TERM"] = "xterm" # nvsm requires
```
Anbang-Hu, line 999: Yes, this was it. I remember I need to set this when ssh executing nvsm for the new version of nvsm.

PR #1038, opened 2020-04-17
expose metrics from gpu reporter

xudifsd:
expose 4 metrics:

* job_booked_gpu_second
* job_idle_gpu_second
* job_non_idle_utils
* user_non_idle_utils

They all have fixed `since` label of 7d, 14d and 31d.

To calculate user's and vc's booked/idle gpu time, just aggregate over job time.

PR #1059, opened 2020-04-26
Remove GPU reporter dependency in RestfulAPI

Gerhut:
Currently RestfulAPI will call gpu reporter for gpu metrics in /GetVC API, this PR will remove these calls and make dashboard directly query Prometheus for those data.

**Require #1038, #1057 to be deployed. If there will be other dashboard deployments before that, do not merge.**

- [x] Unit tests
- [ ] Legacy Cluster Status page support: removed these columns


xudifsd:
so, gpu reporter can remove that gpu_idle endpoint after this?

Gerhut:
> so, gpu reporter can remove that gpu_idle endpoint after this?

If there are no other use, yes.

There will be no direct dependency between gpu reporter and dashboard / restfulapi. Dashboard will (weakly) depends on Prometheus (proxied by Grafana to support https) as it was.

xudifsd:
great

PR #1068, opened 2020-04-30
[Repair Manager] update repairmanager base image

debbie-alaine:
using python base image instead of ubuntu to decrease image size

@xudifsd @Anbang-Hu @YinYangOfDao


debbie-alaine:
After syncing with @xudifsd , we realized ubuntu image is actually smaller

PR #1069, opened 2020-05-01
[RepairManager] improve repairmanager docker images

debbie-alaine:
Currently ubuntu base image is smaller than python base image. 
This Dockerfile change decreases image size from 965MB to 173MB.

@xudifsd @Anbang-Hu @YinYangOfDao

### DIFF
--- a/src/docker-images/RepairManagerAgent/Dockerfile
+++ b/src/docker-images/RepairManagerAgent/Dockerfile
@@ -7,6 +7,8 @@ RUN apt-get update && apt-get --no-install-recommends install -y \
     systemd-sysv
 
 RUN pip3 install wheel setuptools requests
+
+# separate line as pyyaml has dependency on setuptools
 RUN pip3 install pyyaml
 
 COPY run.sh /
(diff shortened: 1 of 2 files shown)


src/docker-images/RepairManagerAgent/Dockerfile
```diff
@@ -1,10 +1,13 @@
...
 MAINTAINER Deborah Sandoval <Deborah.Sandoval@microsoft.com>
 
 RUN apt-get update && apt-get --no-install-recommends install -y \
+    python3.7 \
+    python3-pip \
     systemd-sysv
 
-RUN pip3 install wheel setuptools requests pyyaml
+RUN pip3 install wheel setuptools requests
+RUN pip3 install pyyaml
```
xudifsd, line 10: Why not merge with above line?

debbie-alaine: pyyaml has a dependency on setuptools and error would occur during docker build

xudifsd: ok, you should add a comment

debbie-alaine: image

PR #1070, opened 2020-05-01
fix nccl slowness

xudifsd:
do not terminate nv-hostengine and disable nvlink related metrics.


Anbang-Hu:
Since we have not seen new issues, let's merge this first.

PR #1077, opened 2020-05-04
Add first version of job insight

Anbang-Hu:
```
ALTER TABLE `jobs` ADD COLUMN `insight` LONGTEXT NULL;
```


src/ClusterBootstrap/services/job-insighter/job-insighter-cronjob.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: batch/v1beta1
+kind: CronJob
```
xudifsd, line 2: Why job instead of long running service?

src/RestAPI/dlwsrestapi.py
```diff
@@ -1015,6 +1015,40 @@ def post(self):
         return generate_response(job_priorities)
 
 
+@api.resource("/JobInsight")
```
xudifsd, line 1018: Maybe just Insight?

src/docker-images/job-insighter/insight.py
```diff
@@ -0,0 +1,562 @@
...
+class Insight(object):
+    def __init__(self, job_util, node_spec):
+        self.job_id = job_util.job_id
+        self.since = job_util.since
+        self.end = job_util.end
+
+        # Resource usage pattern from job_util
+        self.job_timespan = job_util.timespan
+        self.num_gpus = len(job_util.gpu)
+        self.idle_gpus = sorted([u for u, v in job_util.gpu.items() if v == 0])
```
xudifsd, line 263: Please do not put computation or function call inside `__init__`. This is not conventional.

src/ClusterBootstrap/services/job-insighter/job-insighter-cronjob.yaml
```diff
@@ -0,0 +1,45 @@
...
+          - name: svccred
+          {% endif %}
+          nodeSelector:
+            job-insighter: active
+          hostNetwork: true
+          restartPolicy: Never
+          tolerations:
+          - key: node-role.kubernetes.io/master
+            effect: NoSchedule
+  schedule: '*/15 * * * *'
```
xudifsd, line 45: What's the behaviour of previous job still running when the next job is about to start? Will two jobs run concurrently? I see `concurrencyPolicy: Forbid`, will next job start running right away after the previous job finished or wait for more time?

src/ClusterBootstrap/services/job-insighter/job-insighter-cronjob.yaml
```diff
@@ -0,0 +1,45 @@
...
+              - 'http://localhost:{{cnf["restfulapiport"]}}'
+              - '--prometheus_url'
+              - 'http://localhost:{{cnf["prometheus"]["port"]}}'
+          {% if cnf["private_docker_registry_username"] %}
+          imagePullSecrets:
+          - name: svccred
+          {% endif %}
+          nodeSelector:
+            job-insighter: active
+          hostNetwork: true
```
xudifsd, line 40: I think we do not need host network?

Anbang-Hu: I am using `localhost` in calls to restfulapi and prometheus. Without host network, it doesn't work.

src/ClusterBootstrap/services/job-insighter/job-insighter-cronjob.yaml
Anbang-Hu: Will change to a service

src/ClusterBootstrap/services/job-insighter/job-insighter-cronjob.yaml
Anbang-Hu: The one second one will not be able to start and be counted as a miss.
https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/

src/RestAPI/dlwsrestapi.py
Anbang-Hu: Changed to `Insight`

src/ClusterBootstrap/services/job-insighter/job-insighter.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: apps/v1
+kind: DaemonSet
```
xudifsd, line 2: If only need one replica, please use Deployment

Anbang-Hu: Changed to Deployment

xudifsd:
I think you also need to add replica count to Deployment like [this](https://github.com/microsoft/DLWorkspace/blob/dltsdev/src/ClusterBootstrap/services/monitor/grafana.yaml#L7)

Anbang-Hu:
> I think you also need to add replica count to Deployment like [this](https://github.com/microsoft/DLWorkspace/blob/dltsdev/src/ClusterBootstrap/services/monitor/grafana.yaml#L7)

If not added, that means 1 by default. Added explicitly.

PR #1078, opened 2020-05-05
add preemptible label to task_* metrics

xudifsd:
Add a `preemptible` label to all `task_*` metrics, with possible value `true` and `false`.

Old jobs already running will have default value `false` regardless their settings.


Anbang-Hu:
Please fix the failed tests before merging.

PR #1079, opened 2020-05-05
Add Infiniband related metrics

Anbang-Hu:
Add below count/perf metrics
1. Infiniband device
2. IPoIB
3. `nv_peer_mem`/GPUDirect

Also add Infiniband related graphs to job status dashboard in Grafana.
image

Note: Infiniband throughput metrics can be scraped only when Mellanox OFED driver is installed on the host machine.


src/docker-images/job-exporter/src/main.py
```diff
@@ -9,6 +9,7 @@
 import faulthandler
 import gc
 import datetime
+import sys
```
Anbang-Hu, line 12: Used in line 219

src/docker-images/job-exporter/src/collector.py
```diff
@@ -717,7 +732,7 @@ def process_one_container(self, container_id, stats, gpu_infos, all_conns,
             if dcgm_infos:
                 for id in gpu_ids:
                     if dcgm_infos.get(id) is None:
-                        contine
+                        continue
```
xudifsd, line 735: good catch

PR #1080, opened 2020-05-05
use user's alias to run inference command

xudifsd:
Also fixed some bugs in deployment.yaml

### DIFF
--- a/src/Jobs_Templete/deployment.yaml.template
+++ b/src/Jobs_Templete/deployment.yaml.template
@@ -37,7 +37,17 @@ spec:
       - name: main
         image: {{ job["image"] }}
         imagePullPolicy: Always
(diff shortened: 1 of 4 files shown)


src/Jobs_Templete/deployment.yaml.template
```diff
@@ -37,7 +37,17 @@ spec:
       - name: main
         image: {{ job["image"] }}
         imagePullPolicy: Always
-        command: ["sh", "-c", "printenv DLTS_LAUNCH_CMD > /pod/job_command.sh ; bash -x /pod/job_command.sh"]
+        command: ["sh", "-c",
+        "
+        printenv DLTS_LAUNCH_CMD > /job_command.sh ;
```
Anbang-Hu, line 42: Just to be cautious, is this tested in dev cluster?

xudifsd: yes

PR #1081, opened 2020-05-06
add preemptible label to aggregated metrics

xudifsd:
Will treat metric without `preemptible` label as with label value `false`. So all previous running job no matter if it is preemptible or not will not be treated as preemptible. Only new preemptible jobs will be count as preemptible jobs.


xudifsd:
Depends on #1078

PR #1084, opened 2020-05-07
add two more metrics to reporters

xudifsd:
Added:
* `vc_allocated_gpu_second`
* `vc_unschedulable_gpu_second`

and aggregate metric:
* `aggregate:vc:vc_allocated_gpu_second:avg`
* `aggregate:vc:vc_unschedulable_gpu_second:avg`

### DIFF
--- a/src/docker-images/gpu-reporter/src/reporter.py
+++ b/src/docker-images/gpu-reporter/src/reporter.py
@@ -158,7 +158,7 @@ class Calculator(object):
         pass
 
     def name(self):
-        pass
+        return self.__class__.__name__
 
     def export(self):
         pass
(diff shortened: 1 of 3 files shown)


src/docker-images/gpu-reporter/src/reporter.py
```diff
@@ -151,7 +153,18 @@ def export(self):
         }
 
 
-class IdlenessCalculator(object):
+class Calculator(object):
+    def observe_metric(self, metric):
+        pass
+
+    def name(self):
+        pass
```
Anbang-Hu, line 161: Not sure if this would be better: `return self.__class__.__name__`

src/docker-images/gpu-reporter/src/reporter.py
```diff
@@ -209,70 +246,94 @@ def export(self):
...
+        calculator.observe_metric(metric)
+
+    result = calculator.export()
+    elapsed = timeit.default_timer() - start
+    logger.info("%s spent %.2fs", calculator.name(), elapsed)
+    return result
+
+
+class Exportable(object):
+    """ object exchanged in AtomicRef, have an export method to generate Metrics
```
Anbang-Hu, line 328: have a `to_metrics` method?

PR #1086, opened 2020-05-08
clean ports

YinYangOfDao:


src/ClusterBootstrap/cloud_init_aztools.py
```diff
@@ -762,6 +656,7 @@ def run_command(command, config, args):
     if command == "deployframework":
         deploy_cluster(config, args)
     if command == "addmachines":
+        os.system('rm -f ' + args.output)
```
Anbang-Hu, line 659: Is this intended change?

YinYangOfDao: yes, it is.

PR #1087, opened 2020-05-08
Only show Infiniband metrics for worker in Grafana dashboard

Anbang-Hu:
#1079 double counts Infiniband throughput for node hosting worker and ps:
image

Restricting to showing only worker metrics:
image

### DIFF
--- a/src/ClusterBootstrap/services/monitor/grafana-config-raw/job-status-dashboard.json
+++ b/src/ClusterBootstrap/services/monitor/grafana-config-raw/job-status-dashboard.json
@@ -47,7 +47,7 @@
   "gnetId": null,
   "graphTooltip": 0,
   "id": null,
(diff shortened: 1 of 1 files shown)


src/ClusterBootstrap/services/monitor/grafana-config-raw/job-status-dashboard.json
```diff
@@ -47,7 +47,7 @@
   "gnetId": null,
   "graphTooltip": 0,
   "id": null,
-  "iteration": 1588881952699,
```
xudifsd, line 50: What's this?

Anbang-Hu: timestamp in epoch millisecond. I guess Grafana uses it in the iteration field.

PR #1089, opened 2020-05-08
add terminating/terminating-timeout phase

xudifsd:
Add `type` label with possible value of `service` or `job` to `k8s_pod_count`.

Also add new possible value to `phase`: `terminating` and `terminating-timeout`.

### DIFF
--- a/src/docker-images/watchdog/src/watchdog.py
+++ b/src/docker-images/watchdog/src/watchdog.py
@@ -76,7 +76,7 @@ def gen_k8s_pod_gauge():
                              labels=[
                                  "service_name", "name", "namespace", "phase",
                                  "host_ip", "initialized", "pod_scheduled",
-                                 "ready"
+                                 "ready", "type"
                              ])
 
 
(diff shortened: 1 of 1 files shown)


src/docker-images/watchdog/src/watchdog.py
```diff
@@ -356,18 +356,37 @@ def parse_pod_item(pod, k8s_pod_gauge, k8s_container_gauge, pods_info,
...
+    else:
+        pod_type = "job"
 
     annotations = walk_json_field_safe(pod, "metadata", "annotations") or {}
     if host_ip != "unscheduled":
         process_service_endpoints(service_name, host_ip, annotations,
                                   service_endpoints)
 
+    deletion_timestamp = walk_json_field_safe(pod, "metadata",
+                                              "deletionTimestamp")
```
Anbang-Hu, line 370: Does older k8s (1.9) have the same fields?

xudifsd: Yes. Just checked.

xudifsd:
enhancement to #1085

PR #1090, opened 2020-05-08
add missing gpuType to inference job

xudifsd:
This will be used in watchdog to count used GPUs.

### DIFF
--- a/src/Jobs_Templete/deployment.yaml.template
+++ b/src/Jobs_Templete/deployment.yaml.template
@@ -20,6 +20,11 @@ spec:
         userName: {{ job["user"] }}
         vcName: {{ job["vcName"] }}
         type: job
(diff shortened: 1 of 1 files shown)


src/Jobs_Templete/deployment.yaml.template
```diff
@@ -20,6 +20,11 @@ spec:
         userName: {{ job["user"] }}
         vcName: {{ job["vcName"] }}
         type: job
+      {% if "gpuType" in job %}
+        {% if job["gpuType"]|length > 0 %}
```
leigaoms, line 24: Where is the definition of "length"?

Anbang-Hu: This is a jinja built-in https://jbmoelker.github.io/jinja-compat-tests/filters/length/

PR #1091, opened 2020-05-09
Storage usage by user

Anbang-Hu:
1. Expose metric `storage_usage_in_bytes_by_user`
2. Support scan point not at `/data/share`
3. Add Grafana storage usage dashboard
image

### DIFF
--- a/src/ClusterBootstrap/params.py
+++ b/src/ClusterBootstrap/params.py
@@ -59,7 +59,7 @@ default_config_parameters = {
     },
 
     "storage-manager": {
-        "port": 9092
+        "port": 9094
     },
 
     "repair-manager": { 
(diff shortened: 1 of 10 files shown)


src/ClusterBootstrap/params.py
```diff
@@ -58,6 +58,10 @@
         }
     },
 
+    "storage-manager": {
+        "port": 9092
```
xudifsd, line 62: gpu-reporter is using 9092

Anbang-Hu: So it's better to use a different port even when they run on different machines?

xudifsd: No. But I think they are running on the same machine? Both are on infra? No?

PR #1095, opened 2020-05-12
add email sender

xudifsd:
The design for this is:

* start a [rabbitmq](https://www.rabbitmq.com/) server outside of CorpNet, maybe deployed along mysql server. I chose rabbitmq because it's easier to deploy than kafka and I once investigated it, [doc](https://gist.github.com/xudifsd/0f2aeff1cf443f7a1b47) in Chinese, its performance should be good for our scenario.
* build a docker image containing code in this PR to serve as both producer and consumer of the queue.
* deploy this image as sidecar of alert manager, redirect all current sending email calls to http call to this server, so the server will put messages to queue.
* also deploy this image inside CorpNet so we can consume data from queue to send out emails.

TODO:

- [x] add authorization to queue, because it will be public server
- [x] deploy along with alert manager
- [x] add scripts to install and deploy rabbitmq and consumer

TODO in other PRs:
* redirect all email sending to this server
* write a webhook for alert manager, because alert manager can only call this via webhook, and need a format transformer

### DIFF
--- a/src/dev-utils/start-rabbit-mq.sh
+++ b/src/dev-utils/start-rabbit-mq.sh
@@ -9,6 +9,7 @@ docker kill rabbitmq
 docker rm rabbitmq
 
 docker run -d \
+    --restart=always \
     --hostname rabbit1 \
     --name rabbitmq \
     --net=host \
--- a/src/dev-utils/start-rabbit-mq.sh
+++ b/src/dev-utils/start-rabbit-mq.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+if [ $# -ne 3 ] ; then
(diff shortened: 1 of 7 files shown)


src/dev-utils/start-rabbit-mq.sh
```diff
@@ -0,0 +1,18 @@
...
+
+if [ $# -ne 3 ] ; then
+    echo Usage: $0 mq_user mq_pass data-dir >&2
+    exit 1
+fi
+
+docker kill rabbitmq
+docker rm rabbitmq
+
+docker run -d \
```
Anbang-Hu, line 11: `--restart=always`

PR #1104, opened 2020-05-13
Dashboard: add storage tab to cluster status

Gerhut:
image


Gerhut:
The table should be ordered by storage bytes, will do it tomo

PR #1110, opened 2020-05-14
Lustre nodes in Kubernetes cluster

Anbang-Hu:
1. Cloud init Kubelet deployment on Lustre nodes (with some caveats on Centos cloud init)
2. Add lustre-exporter
3. Lustre dashboard in Grafana
image

TODO:
- The open source lustre-exporter does not seem to have all the metrics we need. I will open a separate PR for lustre monitoring.


src/ClusterBootstrap/services/monitor/lustre-exporter.yaml
```diff
@@ -0,0 +1,56 @@
...
+      annotations:
+        prometheus.io/scrape: "true"
+        prometheus.io/path: "/metrics"
+        prometheus.io/port: "{{cnf["lustre-exporter"]["port"]}}"
+      name: lustre-exporter
+    spec:
+      nodeSelector:
+        lustre: active
+      containers:
+      - image: dlws/lustre-exporter:v2.0.0
```
xudifsd, line 23: why this image? I think you build the image below

Anbang-Hu: will change to the configurable to avoid breaking prod when there is change

src/ClusterBootstrap/cloud_init_deploy.py
```diff
@@ -264,7 +264,15 @@ def load_default_config(config):
...
 
-        "lustre_server": ["./deploy/storage/auto_share/setup_lustre.sh"]}
+        "lustre_server": [
+            "./deploy/kubelet/" + config["preworkerdeploymentscript"],
+            "./deploy/kubelet/" + config["postworkerdeploymentscript"],
+            "./scripts/cloud_init_lustre.sh",
+            "./scripts/prepare_lustre_centos.sh",
+            "./deploy/storage/auto_share/setup_lustre.sh",
+            "./deploy/cloud-config/lustre.kubelet.service.template"
+        ]
```
YinYangOfDao, line 274: please add these to filemap, or I would refactor all of these later.

src/ClusterBootstrap/scripts/cloud_init_lustre.sh
```diff
@@ -0,0 +1,23 @@
...
+#bash ./pre-worker-deploy.sh
+#source ../boot.env
+#./cloud_init_mkdir_and_cp.py -p file_map.yaml -u $USER -m $MOD_2_CP
+#
+#./render_env_vars.sh kubelet_worker/deploy/kubelet/options.env.template /etc/flannel/options.env ETCD_ENDPOINTS
+#./render_env_vars.sh kubelet_worker/deploy/kubelet/worker-kubeconfig.yaml.template /etc/kubernetes/worker-kubeconfig.yaml KUBE_API_SERVER
+#./render_env_vars.sh lustre.kubelet.service.template /etc/systemd/system/kubelet.service KUBE_LABELS
+#
+#bash ./post-worker-deploy.sh
+#-----------------------------------------------------------------------------
```
YinYangOfDao, line 20: do you want the commented lines? if not, this file should be deleted

src/ClusterBootstrap/template/storage/auto_share/setup_lustre.sh
```diff
@@ -40,6 +40,11 @@ touch {{cnf["folder_auto_share"]}}/lustre_setup_finished
 chown -R {{cnf["lustre_user"]}}:{{cnf["lustre_user"]}} {{cnf["folder_auto_share"]}}
 # we further setup service that examine 2 token: when lustre_setup_finished exist, lustre_nodes_locally_mounted doesn't, execute following script
 source ../boot.env
-python ./cloud_init_mkdir_and_cp.py -p file_map.yaml -u $USER -m $MOD_2_CP
+
+# file system operations in Centos cloud init is extremely slow - a few to
```
YinYangOfDao, line 44: to confirm, is it cloudinit that make it slow?

src/ClusterBootstrap/scripts/cloud_init_lustre.sh
Anbang-Hu: From deployment perspective, these commands should better be here. It is due to the perf issue that I choose to move them to `lustre_mdt_or_oss.sh`. I do have a `TODO` comment saying "look into the root cause to the slowness".

src/ClusterBootstrap/template/storage/auto_share/setup_lustre.sh
Anbang-Hu: Correct. If I start a VM and manually execute the command, the latency is good.

Anbang-Hu: We should keep this in mind in case there are other roles requiring centos in the future.

src/ClusterBootstrap/cloud_init_deploy.py
Anbang-Hu: Since adding to filemap also applies to worker, infra, etc., let's refactor all these in a separate PR.

PR #1113, opened 2020-05-14
let repair manager use email-sender

xudifsd:
Use `EmailMessage` to represent email message internally, and convert needed format when actually sending, if configured `email_sender_url` will convert to json and send to email sender. If not will convert to `MIMEMultipart` and send via smtp.

Needs following config:
```
repair-manager:
  email_sender_url: http://localhost:9095/put_email
  default_recipients: abc@microsoft.com
```


xudifsd:
Hold this PR, find another way to do this.

PR #1120, opened 2020-05-19
Lustre metrics in job exporter

Anbang-Hu:
[lustre-exporter](https://github.com/HewlettPackard/lustre_exporter) does not seem to give all the info we want when deployed as a docker. Instead of modifying the code, a more manageable way for us is to add the relevant scraping in job-exporter.

To start with, the metrics of great interest are
- Storage size (capacity and usage for both bytes and inodes)
- Storage throughput (bandwidth and IOPs)

image


src/docker-images/job-exporter/src/lustre.py
```diff
@@ -0,0 +1,311 @@
...
+    # create                    919 samples [reqs]
+    # statfs                    49389 samples [reqs]
+    # get_info                  1 samples [reqs]
+    # set_info                  60 samples [reqs]
+    # ...
+    #
+    # 2.
+    # health_check=healthy
+    # TODO: Find the proper regex to do this.
+    groups = collections.defaultdict(lambda: list())
```
xudifsd, line 186: `lambda: []`

src/docker-images/job-exporter/src/lustre.py
```diff
@@ -0,0 +1,311 @@
...
+    if lctl_get_param("mds.MDS.uuid"):
+        return "mds"
+    elif lctl_get_param("ost.OSS.uuid"):
+        return "oss"
+    else:
+        return "client"
+
+
+def group_content(content):
+    # Contents look like
```
xudifsd, line 155: You can write a unit test with sample output

src/docker-images/job-exporter/src/lustre.py
```diff
@@ -0,0 +1,311 @@
...
+
+    lustre_metrics = []
+    for role, metric_tuples in metric_mapping.items():
+        for name, pattern, desc in metric_tuples:
+            lustre_metrics.append(
+                LustreMetric(name, pattern, desc, server, role))
+
+    # Different metrics can be derived from the same lctl param pattern call
+    # pattern -> [metric]
+    pattern_to_lmetrics = collections.defaultdict(lambda: list())
```
xudifsd, line 284: use literal value whenever possible

PR #1122, opened 2020-05-19
Use markdown in insight diagnostics

Anbang-Hu:


xudifsd:
Why not just output html in insight?

PR #1132, opened 2020-05-22
add up.sh for upgrading

xudifsd:
For upgrading the clusters.

Still need human check, so I added watch command at the end of the script to display the check infor.

I think we can refine this with fresh git clone. So it can be executed in some dev-box for upgrading only.

### DIFF
--- a/src/dev-utils/up.sh
+++ b/src/dev-utils/up.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+set -ex
+
+# This script assume code repo and config repo are clean
+
+CODE_BASE_DIR="/mnt/dev/dlts"
+CODE_UPSTREAM_REMOTE="upstream"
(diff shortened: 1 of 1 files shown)


src/dev-utils/up.sh
```diff
@@ -0,0 +1,53 @@
...
+./ctl.py restorefromdir $CONFIG_DIR/$cluster_name
+
+# start to deploy
+
+for i in `echo restfulapi storagemanager repairmanager dashboard`
+do
+    ./ctl.py svc configupdate $i
+done
+
+./ctl.py kubectl label node `grep infra01: status.yaml  | sed 's/ *:*//g'` job-insighter=active
```
Anbang-Hu, line 38: infra node name does not necessarily contain `infra01`, but we can refine later.

xudifsd: ok

PR #1133, opened 2020-05-22
Logging: fix incorrect time-based blob ordering

Gerhut:
Since the `last_modified` field of blob is in seconds, it's found that the previous blob has the same last_modified as the next one. That makes the sort useless.

### DIFF
--- a/src/docker-images/azure-blob-adapter/wsgi.py
+++ b/src/docker-images/azure-blob-adapter/wsgi.py
@@ -21,6 +21,13 @@ container_name = environ['AZURE_STORAGE_CONTAINER_NAME']
 append_blob_service = AppendBlobService(connection_string=connection_string)
 
 
+def _get_blob_index(blob_name):
+    try:
+        return int(blob_name.split('.', 2)[2])
+    except (IndexError, ValueError):
+        return 0
+
+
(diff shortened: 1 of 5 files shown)


src/docker-images/azure-blob-adapter/wsgi.py
```diff
@@ -21,6 +21,13 @@
 append_blob_service = AppendBlobService(connection_string=connection_string)
 
 
+def _get_blob_index(blob_name):
+    try:
+        return int(blob_name.split('.', 1)[1])
```
Anbang-Hu, line 26: I remember the `blob_name` has format `jobs.<job id>.<idx>`? If so, should this be `int(blob_name.split('.')[-1])`?

Gerhut: Thanks for reminding, my bad. Fixed with `[2]`.

Retrieving the 3rd is for triggering the `IndexError` in case of there are only 2 items in the split list (the first blob is named `jobs.<job-id>`).

xudifsd: Please add a test case for this.

Gerhut:
Done unit tests, based on #1160

PR #1136, opened 2020-05-26
Dashboard: add download Icon to the console tab

Gerhut:
Depends on #1127

image


Gerhut:
Need additional unit tests to keep the test coverage

Anbang-Hu:
Is the download button per pod?

PR #1137, opened 2020-05-26
remove cleaning old job code

xudifsd:
The deleting query takes too long because we didn't index `lastUpdated` column of jobs table.

We can later do this by either index it or query all jobId out and delete them.

This long delete query will result out update job failed:
```
2020-05-26 02:18:53,844 - INFO - job_manager.py:304 - Job status: 79f79edd-96b6-4666-b380-67f9fa5e89f5 Succeeded
2020-05-26 02:19:46,377 - ERROR - MySQLDataHandler.py:1281 - failed to UpdateJobTextFields conditions {'jobId': '79f79e
dd-96b6-4666-b380-67f9fa5e89f5'}, data {'jobStatus': 'finished', 'jobStatusDetail': 'x'}
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/mysql/connector/connection_cext.py", line 489, in cmd_query
    raw_as_string=raw_as_string)
_mysql_connector.MySQLInterfaceError: Lock wait timeout exceeded; try restarting transaction

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "../utils/MySQLDataHandler.py", line 1275, in UpdateJobTextFields
    cursor.execute(sql, values)
  File "/usr/local/lib/python3.5/dist-packages/mysql/connector/cursor_cext.py", line 266, in execute
    raw_as_string=self._raw_as_string)
  File "/usr/local/lib/python3.5/dist-packages/mysql/connector/connection_cext.py", line 492, in cmd_query
    sqlstate=exc.sqlstate)
mysql.connector.errors.DatabaseError: 1205 (HY000): Lock wait timeout exceeded; try restarting transaction
```

```
---TRANSACTION 1604714425, ACTIVE 811 sec fetching rows
mysql tables in use 1, locked 1
54529 lock struct(s), heap size 4874448, 248928 row lock(s), undo log entries 51735
MySQL thread id 1139072470, OS thread handle 140373082511104, query id 9418009966 xxx updating
DELETE FROM jobs WHERE lastUpdated < NOW() - INTERVAL 180 DAY AND `jobStatus` IN ('finished','failed','killed','error')
---TRANSACTION 1604702461, ACTIVE 924 sec
49342 lock struct(s), heap size 4415696, 314854 row lock(s), undo log entries 3662
MySQL thread id 1139051414, OS thread handle 140373060392704, query id 9417840122 xxx
```

PR #1138, opened 2020-05-26
remove cleaning old job code

xudifsd:
cherry pick of #1137


xudifsd:
I'm not sure if adding index to lastUpdated will speed up this queries. May need investigation. An alternative is query them and delete them in small batch.

PR #1141, opened 2020-05-29
Dashboard/home/fix

Gerhut:
- [x] Fix percentage format
- [x] Fix mountpoint mapping
- [x] Get directory text boxes back
- [x] Revert mountpoint style


Gerhut:
> Please help change `~` back to `/data`
> image
> Please also help bring back this rule:
> 
> * Show orange bar when used >= 80%
> * Show red bar when used >= 90%
>   image

done

PR #1144, opened 2020-05-30
make job-priority as global default

xudifsd:
Without default priority, jobs created by previous version will be preempted by jobs created by new version. Because k8s set 0 as default value.

### DIFF
--- a/src/docker-images/RestfulAPI/runScheduler.sh
+++ b/src/docker-images/RestfulAPI/runScheduler.sh
@@ -10,7 +10,7 @@ cur_pgid=`ps -o '%r' $$ | tail -n 1 | sed "s/ //g"`
 
 trap _broadcast SIGTERM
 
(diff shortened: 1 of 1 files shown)


Anbang-Hu:
Is this fully tested on dev1 cluster? I understand you tested adding a new priority class, just want to make sure modifying the job priority also serves the same purpose.

xudifsd:
> Is this fully tested on dev1 cluster? I understand you tested adding a new priority class, just want to make sure modifying the job priority also serves the same purpose.

Yes, I've tested several senario:

* Do not have default priority: previous scheduled pods with no priority class will be preempted.
* Have default priority with value 1000 but with different priority class name(like case in RR1): previous scheduled pods created with default priority class in effect will not be preempted.

To deploy this patch to RR1 and 32GB, will need to delete default priority class, and redeploy job manager.

PR #1147, opened 2020-06-02
support set max running time for job

xudifsd:
DRI can set VC with parameter like:
```json
{"P40":{"num_gpu_per_node":4}, "admin": {"job_max_time_second": 259200}}
```

Then restfulapi will check this field and add `maxTimeSec` to job parameter. job manager will check this field and compare with `lastUpdated` time, if time spend in Running exceed max time, the job manager will kill the job and replace `errorMsg` with some meaningful words.

This PR also expose an API so that only VC admin can call: `/JobMaxTime`. So that VC admin can adjust this value.

@Gerhut please help change front end:

* add a field display job's max time value by checking `maxTimeSec` field of job's parameter, if it's not exist, then maybe display +∞, otherwise maybe convert it into hour or day for easy display
* allow VC admin adjust this value, and call `/JobMaxTime` API. If this value is too small, then the job may be killed immediately.

### DIFF
--- a/src/ClusterManager/job_manager.py
+++ b/src/ClusterManager/job_manager.py
@@ -364,8 +364,9 @@ def UpdateJobStatus(redis_conn,
         params = json.loads(base64decode(job["jobParams"]))
         max_time = params.get("maxTimeSec")
         if type(max_time) != int:
-            logger.info("unknown maxTimeSec %s for job %s", max_time,
-                        job["jobId"])
+            if max_time is not None:
+                logger.info("unknown maxTimeSec %s for job %s", max_time,
+                            job["jobId"])
         else:
             max_time = int(max_time)
             start_time = int(datetime.datetime.timestamp(job["lastUpdated"]))
--- a/src/ClusterManager/job_manager.py
+++ b/src/ClusterManager/job_manager.py
@@ -322,7 +322,8 @@ def UpdateJobStatus(redis_conn,
 
         dataFields = {
             "jobStatusDetail": b64encode(json.dumps(detail)),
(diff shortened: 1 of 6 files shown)


xudifsd:
I see dashboard in int is not displaying `errorMsg` value, not sure why.

src/ClusterManager/test/test_regular_job.py
```diff
@@ -947,3 +947,21 @@ def test_do_not_starve_preemptible_job(args):
...
+
+    with utils.run_job(args.rest, job_spec) as job:
+        state = job.block_until_state_not_in(
+            {"unapproved", "queued", "scheduling"})
+        assert state == "running"
+
+        resp = utils.set_job_max_time(args.rest, args.email, job.jid, 1)
+        assert resp.status_code == 200
+
+        assert state == "running"
```
Anbang-Hu, line 965: This looks redundant.

src/ClusterManager/job_manager.py
```diff
@@ -355,6 +359,27 @@ def UpdateJobStatus(redis_conn,
                     notify.new_job_state_change_message(job["userName"],
                                                         job["jobId"],
                                                         result.strip()))
+            job["lastUpdated"] = last_updated
+
+        params = json.loads(base64decode(job["jobParams"]))
+        max_time = params.get("maxTimeSec")
+        if type(max_time) != int:
+            logger.info("unknown maxTimeSec %s for job %s", max_time,
```
Anbang-Hu, line 367: For a cluster with no max time, this could be a lot of logs.

PR #1150, opened 2020-06-02
make up.sh easier to customize

xudifsd:
to customize upgrade, just comment out unnecessary components and the script helps automate the whole thing.

### DIFF
--- a/src/dev-utils/up.sh
+++ b/src/dev-utils/up.sh
@@ -104,7 +104,6 @@ job-insighter
 reaper
 repairmanager
 repairmanageragent
-repairmanageretcd
 storagemanager
 user-synchronizer
 watchdog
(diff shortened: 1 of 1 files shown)


src/dev-utils/up.sh
```diff
@@ -9,45 +9,119 @@ CODE_UPSTREAM_REMOTE="upstream"
...
+
+restore_config
+
+update_service_config `cat <(cat << EOF
+restfulapi
+storagemanager
+repairmanager
+dashboard
+EOF
+) | grep -v '#' `
```
Anbang-Hu, line 92: Why do we need `grep -v '#'`?

xudifsd: I want to use this script in almost all upgrade. For example, if we only need to upgrade jobmanager, then we can comment out all unnessary steps and components.

PR #1151, opened 2020-06-02
make stop time shorter for repair/storage manager

xudifsd:
Previous version ignore SIGTERM which will result slower restart time.

Ref #1139


src/ClusterBootstrap/params.py
```diff
@@ -112,7 +112,6 @@
     # StorageManager mapping
     "storagemanager": "storagemanager",
     "repairmanager": "repairmanager",
-    "repairmanageretcd": "repairmanageretcd",
```
Anbang-Hu, line 115: I'm pretty sure this is unintentional. Please add repairmanageretct code back.

src/ClusterBootstrap/services/repairmanager/repairmanager.yaml
```diff
@@ -45,8 +45,8 @@ spec:
         - mountPath: /etc/RepairManager
           name: repairmanagerdata
       - name: etcd
-        image: {{cnf["worker-dockerregistry"]}}/{{cnf["dockerprefix"]}}/{{cnf["repairmanageretcd"]}}:{{cnf["dockertag"]}}
-        command: ["/run.sh"]
+        image: quay.io/coreos/etcd:v3.3.22
+        command: ["etcd", "--config-file", "/etc/RepairManager/config/etcd.conf.yaml"]
```
Anbang-Hu, line 49: This was not working all along?

PR #1157, opened 2020-06-04
restrict how many gpus can be used in interactive jobs

xudifsd:
VC admin can add `admin` -> `interactive_limit` with an int value to restrict how many gpu can be used in interactive jobs.

We use endpoints to check if a job is interactive or not: if a job only opened a tensorflow or none endpoint, it will be deemed as non-interactive job, if a job opened other endpoints, it will be deemed as interactive job.

Also exposed following metrics:

* `job_manager_vc_{cpu,gpu,memory}_{capacity,used,preemptable_used,available,unschedulable,reserved}`
* `job_manager_vc_gpu_interactive_used`

Later TODO:

* expose if job is interactive of not

@Gerhut Impact on front end: post to `/endpoints` may return 403 along with plain text reason, you can display this reason to user directly.

### DIFF
--- a/src/utils/JobRestAPIUtils.py
+++ b/src/utils/JobRestAPIUtils.py
@@ -1342,28 +1342,25 @@ def UpdateEndpoints(userName, jobId, requested_endpoints, interactive_ports):
                 walk_json(dataHandler.GetVC(job["vcName"]),
                           "metadata",
                           default="{}"))
-            interactive_ratio = float(
-                walk_json(vc_meta, "admin", "interactive_ratio", default=100.0))
+            interactive_limit = walk_json(vc_meta, "admin", "interactive_limit")
 
-            if interactive_ratio == 100.0:
+            if interactive_limit is None:
                 allowed = True
             else:
-                gpu_type = job_params["sku"]
+                interactive_limit = int(interactive_limit)
+                sku = job_params["sku"]
 
                 status, _ = dataHandler.GetClusterStatus()
                 vc_status = walk_json(status, "vc_statuses", job["vcName"])
-                required = Gpu({gpu_type: gpu_request})
+                required = Gpu({sku: gpu_request})
                 interactive_used = Gpu(
                     walk_json(vc_status, "gpu_interactive_used", default={}))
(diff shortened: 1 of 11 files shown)


src/utils/JobRestAPIUtils.py
```diff
@@ -1300,9 +1321,53 @@ def UpdateEndpoints(userName, jobId, requested_endpoints, interactive_ports):
...
+                          default="{}"))
+            interactive_limit = walk_json(vc_meta, "admin", "interactive_limit")
+
+            if interactive_limit is None:
+                allowed = True
+            else:
+                interactive_limit = int(interactive_limit)
+                sku = job_params["sku"]
+
+                status, _ = dataHandler.GetClusterStatus()
```
Anbang-Hu, line 1353: Not sure if this will become too heavy. We can optimize later.

src/utils/JobRestAPIUtils.py
```diff
@@ -1300,9 +1321,53 @@ def UpdateEndpoints(userName, jobId, requested_endpoints, interactive_ports):
...
+                vc_status = walk_json(status, "vc_statuses", job["vcName"])
+                required = Gpu({sku: gpu_request})
+                interactive_used = Gpu(
+                    walk_json(vc_status, "gpu_interactive_used", default={}))
+                if required + interactive_used <= interactive_limit:
+                    allowed = True
+                else:
+                    allowed = False
+                    reason = "Can not open interactive port as required %s GPU and %s GPUs are already in interactive usage, exceed limit of %s, you can ask admin to enable this endpoint if you do want" % (
```
Anbang-Hu, line 1362:
```
"Cannot open interactive endpoint. There are already %s interactive GPUs in the VC, request another %s interactive GPUs exceeds the limit of %s. You can ask admin to enable the interactive endpoint." % (interactive_used, gpu_request, interactive_limit)
```

src/utils/JobRestAPIUtils.py
xudifsd: Yes. I think this need to be optimized.

PR #1158, opened 2020-06-04
Increase nfsd count to reduce the possibility of NFS hanging

Anbang-Hu:


xudifsd:
Are you sure this can help?

Anbang-Hu:
> Are you sure this can help?

At least 4 times is a better start. nfsd mostly handles IOs.

PR #1142, opened 2020-05-29
change scheduling policy, do not starve big job

xudifsd:
Do not try to schedule smaller job if one bigger job is stuck in getting resource.

PR #1168, opened 2020-06-10
vc wide block instead of cluster wide

xudifsd:
solve a bug introduced by #1142 

Block scheduling within VC, instead of cluster wide.

### DIFF
--- a/src/ClusterManager/job_manager.py
+++ b/src/ClusterManager/job_manager.py
@@ -560,7 +560,7 @@ def get_jobs_info(jobs, cluster_schedulable, vc_schedulables):
             vc_schedulable = vc_schedulables.get(vc_name)
             if vc_schedulable is None:
                 logger.warning(
-                    "vc %s is not exist as provided by %s, ignore this job",
+                    "vc %s does not exist as provided by %s, ignore this job",
                     vc_name, job_id)
                 continue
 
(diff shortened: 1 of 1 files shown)


src/ClusterManager/job_manager.py
```diff
@@ -556,6 +556,22 @@ def get_jobs_info(jobs):
             job_res = get_resource_params_from_job_params(job_params)
             job_resource = ClusterResource(params=job_res)
 
+            vc_name = job["vcName"]
+            vc_schedulable = vc_schedulables.get(vc_name)
+            if vc_schedulable is None:
+                logger.warning(
+                    "vc %s is not exist as provided by %s, ignore this job",
```
Anbang-Hu, line 563: `vc %s does not exist ...`

src/ClusterManager/job_manager.py
```diff
@@ -601,7 +617,7 @@ def get_jobs_info(jobs):
 
 def mark_schedulable_non_preemptable_jobs(jobs_info, cluster_schedulable,
                                           vc_schedulables):
-    stop_scheduling = None
+    stop_schedulings = {} # vc_name -> the first blocking job of this vc
```
Anbang-Hu, line 620: We need a way to configure this policy for a specific VC.

xudifsd: yes, working

PR #1169, opened 2020-06-10
allow user provide max retry count from restfulapi

xudifsd:
Support `max_retry_count` in job params. Should be an int or int value.


xudifsd:
@Gerhut Could you help add this support in dashboard? Maybe in Advanced setting.

PR #1170, opened 2020-06-10
kill with description

xudifsd:
Allow extra parameter `desc` for `KillJob`. Will put this message to `errorMsg`.

### DIFF
--- a/src/docker-images/reaper/main.py
+++ b/src/docker-images/reaper/main.py
@@ -33,10 +33,16 @@ def kill():
             logger.info("processing alert of %s", alert)
             if not dry_run:
                 job_name = alert["labels"]["job_name"]
+                idle_hour_desc = alert.get("annotations",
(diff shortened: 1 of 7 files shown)


src/docker-images/reaper/main.py
```diff
@@ -33,7 +33,11 @@ def kill():
             logger.info("processing alert of %s", alert)
             if not dry_run:
                 job_name = alert["labels"]["job_name"]
-                params = {"jobId": job_name, "userName": "Administrator"}
+                params = {
+                    "jobId": job_name,
+                    "userName": "Administrator",
+                    "desc": "killed because the job is being idle too long"
```
Anbang-Hu, line 39: Is it possible to expose the max GPU idle hours to users in the message?

xudifsd: Yes, will need to add label to alert and receive label here.

PR #1172, opened 2020-06-11
Dashboard: add timeout field

Gerhut:
Merge priority column with the actual field before do the actual code, in order to fork to the timeout field.

image


xudifsd:
Front end support for #1147

xudifsd:
Is this finished or not? Do you support user input "1d" and convert to "86400" second? Also display this instead of in second.

Anbang-Hu:
> Is this finished or not? Do you support user input "1d" and convert to "86400" second? Also display this instead of in second.

Let's use hours. We use GPU hours as the unit for GPU compute consumption by users.

Gerhut:
Deployed to int

PR #1174, opened 2020-06-15
Dashboard/job insights

Gerhut:
@Anbang-Hu sorry to revert your markdown efforts, I found it should be listed and deal message by message.

If necessary I will split the revert commit into a separated PR.

image

- Currently the job insights "GPU idle" and "need more metrics" seems have different message levels (warn and info). They may have different styles for user if we can do that.
- It's better to give an action to user for each message, like "kill"(for gpu idle), "support"(maybe by default) or "dismiss"(cannot be persistent currently)

Expected fields:
- level
- message
- action

- [ ] drop the revert commit and rebase on the latest dev branch

### DIFF
--- a/src/dashboard/src/pages/Job/Insight.tsx
+++ b/src/dashboard/src/pages/Job/Insight.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import {
   FunctionComponent,
-  SyntheticEvent,
+  memo,
   useCallback,
   useContext,
   useMemo,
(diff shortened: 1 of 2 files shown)


src/docker-images/job-insighter/insight.py
```diff
@@ -506,9 +508,9 @@ def upload_insights(insights, restful_url, dry_run):
             logger.exception("failed to upload insight for %s", job_id)
 
 
-def run(prometheus_url, mins_ago, restful_url, dry_run):
+def run(prometheus_url, hours_ago, restful_url, dry_run):
```
Anbang-Hu: Can you keep the change regarding `minutes`?

Gerhut:
image

Deployed to int, seems the content is too much

Anbang-Hu:
> image
> 
> Deployed to int, seems the content is too much

Let's use a collapseable section showing the first line with ellipses by default.

In addition, job insight should only show for running jobs.

Gerhut:
image

Gerhut:
image

---

image

Gerhut:
changed to `SHOW ALL` to keep user from reminding of collapse from expand

PR #1182, opened 2020-06-16
Repair cycle of worker nodes

Anbang-Hu:
image
image

TODO:
1. Fix the case when metrics have not been populated in Prometheus for a node in `AFTER_REPAIR`.
2. Expose repair metrics in Prometheus (#nodes in different repair state, etc.)
3. Email handler
4. NotReady node needs to be marked as unschedulable (need more thought on this)
5. Allow an unschedulable node marked by dev to not go into repair cycle.
6. Agent should keep repair if the repair fails.

### DIFF
--- a/src/ClusterBootstrap/services/repairmanager/repairmanager.yaml
+++ b/src/ClusterBootstrap/services/repairmanager/repairmanager.yaml
@@ -120,6 +120,8 @@ spec:
         volumeMounts:
         - mountPath: /var/log/repairmanager
           name: log
+        - mountPath: /host-fs
+          name: host-fs
         ports:
         - name: agent
           containerPort: {{cnf["repair-manager"]["agent-port"]}}
@@ -131,6 +133,9 @@ spec:
       - name: log
         hostPath:
           path: /var/log/repairmanager
+      - name: host-fs
+        hostPath:
+          path: /
       tolerations:
       - key: CriticalAddonsOnly
         operator: Exists
--- a/src/ClusterBootstrap/services/repairmanager/repairmanager.yaml
+++ b/src/ClusterBootstrap/services/repairmanager/repairmanager.yaml
@@ -40,7 +40,6 @@ spec:
           - '--interval=30'
           - '--port={{cnf["repair-manager"]["port"]}}'
           - '--agent_port={{cnf["repair-manager"]["agent-port"]}}'
-          - '--dry_run'
         imagePullPolicy: Always
         volumeMounts:
         - mountPath: /etc/kubernetes/ssl
@@ -116,7 +115,6 @@ spec:
           - '--service=repairmanageragent'
           - '--log=/var/log/repairmanager'
           - '--agent_port={{cnf["repair-manager"]["agent-port"]}}'
-          - '--dry_run'
         imagePullPolicy: Always
         volumeMounts:
         - mountPath: /var/log/repairmanager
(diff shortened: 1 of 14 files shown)


src/RepairManager/src/rule.py
```diff
@@ -0,0 +1,421 @@
...
+
+    Returns:
+        None
+    """
+    if not isinstance(commands, list):
+        commands = [commands]
+    for command in commands:
+        command = CHROOT_HOST_FS + command
+        logger.info("Executing command: %s", command)
+        os.system(command)
```
xudifsd, line 33: Maybe use `Popen` to capture stdout and stderr and log. Use os.system directly will put them to current process stdout and stderr, which might lost when docker restart. And Popen also provide timeout.

Also, maybe the function name should be `exec_commands`. Should make it accept list only. And command should better be in form of list, the string format will invoke a sh to interpret the command.

I think it's better to define two functions:

* exec_command
* exec_commands

The first accept a list, the second accept a list of list.

With above change:
* `exec_command("reboot -f")` -> `exec_command(["reboot", "-f"])`
* `exec_command(["nvsm dump health", "sync", "reboot -f"])` -> `exec_commands([["nvsm", "dump", "health"], ["sync"], ["reboot", "-f"]])`

src/RepairManager/src/repairmanager.py
```diff
@@ -0,0 +1,421 @@
...
+    def update(self, node):
+        """Defines the status change of node.
+
+        IN_SERVICE --check_health True--> IN_SERVICE
+                  \--check_health False--> OUT_OF_POOL
+
+        OUT_OF_POOL --prepare True--> READ_FOR_REPAIR
+                   \--prepare False--> OUT_OF_POOL
+
+        READ_FOR_REPAIR --send_repair_signal True--> IN_REPAIR
```
xudifsd, line 97: `send_repair_request `

src/RepairManager/src/rule.py
```diff
@@ -0,0 +1,421 @@
...
+    def __init__(self, metrics, stat="avg", interval="5m"):
+        """Constructs a Rule instance.
+
+        Args:
+            metrics: A string or a list of metrics.
+            stat: avg, min, max, etc. supported by <stat>_over_time query
+                in Prometheus.
+            interval: The look-back time interval for stat.
+        """
+        self.metrics = metrics if isinstance(metrics, list) else [metrics]
```
xudifsd, line 79: I think it's better to make sure the caller passes in list.

src/ClusterBootstrap/services/repairmanager/repairmanager.yaml
```diff
@@ -62,29 +64,83 @@ spec:
...
+      - name: repairmanageragent
+        image: {{cnf["worker-dockerregistry"]}}/{{cnf["dockerprefix"]}}/{{cnf["repairmanager"]}}:{{cnf["dockertag"]}}
+        securityContext:
+          capabilities:
+            add: ["SYS_BOOT"]
+        command:
+          - "python3"
+          - "/repairmanager/main.py"
+        args:
+          - '--service=repairmanageragent'
```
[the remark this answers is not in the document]
xudifsd: Can leave it as it is.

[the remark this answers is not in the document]
Anbang-Hu: Split into `repairmanager.py` and `agent.py`

src/RepairManager/src/repairmanager.py
Anbang-Hu: Changed

src/RepairManager/src/rule.py
Anbang-Hu: Different command may have different timeout. Implemented `exec_cmd(command)` where `command` is a list.

PR #1188, opened 2020-06-17
Dashboard: open samba folder with a click

Gerhut:
## Usage

1. Download [this PowerShell script](https://github.com/Gerhut/browse/blob/master/browse.ps1) and put it in a stable location.
2. Run the script once to register the `browse:` URL protocol to open the rest part in `explorer.exe`. It will automatically enter Administrator mode if not.
3. Now it is able to open the samba directory in explorer with a click of the button next to the working directory, which navigates to `browse:file://samba/` -> `explorer.exe file://samba/`.

image

### DIFF
--- a/src/dashboard/src/pages/Home/ClusterCard/DirectoryContent.tsx
+++ b/src/dashboard/src/pages/Home/ClusterCard/DirectoryContent.tsx
@@ -15,7 +15,7 @@ import {
   Tooltip,
 } from '@material-ui/core';
 import {
-  FileCopyRounded,
+  OpenInNew,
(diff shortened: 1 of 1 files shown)


Anbang-Hu:
This is a really good feature. Can we somehow embed the Powershell script on dashboard? When a user first clicks on the Samba path, it is better to error out with some helpful instructions on how to use this feature.

Gerhut:
It's hard to decide whether the user has installed the support of the URL protocol due to security concerns. So we need to consider other UX workarounds to support both user installed and not installed the support.

PR #1191, opened 2020-06-18
Dashboard/endpoint/fix

Gerhut:
- Transfer restfulapi error message when 403
- Use vanilla fetch to get the error message
- reset the switch when fails to enable

### DIFF
--- a/src/dashboard/src/pages/Job/Tabs/Endpoints.tsx
+++ b/src/dashboard/src/pages/Job/Tabs/Endpoints.tsx
@@ -112,15 +112,19 @@ const EndpointsController: FunctionComponent<{ endpoints: any[] }> = ({ endpoint
   const tensorboard = useMemo(() => {
     return endpoints.some((endpoint) => endpoint.name === 'tensorboard');
   }, [endpoints]);
-  const { post } =
-    useFetch(`/api/clusters/${clusterId}/jobs/${jobId}/endpoints`,
-    [clusterId, jobId]);
   const portInput = useRef<HTMLInputElement>();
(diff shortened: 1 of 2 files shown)


Anbang-Hu:
I don't see the feature working properly in the cluster that has interactive limit deployed. I am unsure if which part goes wrong. @xudifsd can you help look into it?

Anbang-Hu:
> I don't see the feature working properly in the cluster that has interactive limit deployed. I am unsure if which part goes wrong. @xudifsd can you help look into it?

Looks like I was in another security group that has admin access to the VC.

src/dashboard/src/pages/Job/Tabs/Endpoints.tsx
```diff
@@ -112,15 +115,19 @@ const EndpointsController: FunctionComponent<{ endpoints: any[] }> = ({ endpoint
...
-  const { post } =
-    useFetch(`/api/clusters/${clusterId}/jobs/${jobId}/endpoints`,
-    [clusterId, jobId]);
   const portInput = useRef<HTMLInputElement>();
   const onChange = useCallback((name: string) => (event: ChangeEvent<{}>, value: boolean) => {
     if (value === false) return;
     enqueueSnackbar(`Enabling ${name}...`);
-    post({
-      endpoints: [name.toLowerCase()]
+    // TODO: revert to use-http after upgrade to 1.0
```
Anbang-Hu, line 122: Why revert after upgrade?

Gerhut:
Backend continues on #1197 
Frontend continues on #1199

PR #1193, opened 2020-06-18
support vc edit vc metadata

xudifsd:
Add two restful api:

* GET `VCMeta`
* POST `VCMeta`

When posting to change, can submit partial meta, and null value with a key will make it null, for example, post `{"job_max_time_second": null}` will turn off this in this vc.

Restful api also validate the meta:
* `job_max_time_second` should be null or int
* `interactive_limit` should be null or int
* `scheduling_policy` should be "RF" or "FIFO"

Ref #1147 #1157 #1177 

Fixed a bug in #1177 .

Also introduce `--dangerous` to integration test, for test case requiring change vc meta, will rollback to old meta, do not test these cases on production clusters.

### DIFF
--- a/src/ClusterManager/test/test_regular_job.py
+++ b/src/ClusterManager/test/test_regular_job.py
@@ -888,7 +888,7 @@ def test_regular_job_no_distributed_system_envs(args):
                     expected_output, output)
 
 
-#@utils.case() TODO setup vc config to test this
+@utils.case(dangerous=True)
 def test_do_not_starve_big_job(args):
     big_job_spec = utils.gen_default_job_description("regular",
                                                      args.email,
@@ -900,27 +900,30 @@ def test_do_not_starve_big_job(args):
                                                        args.uid, args.vc)
     small_job_spec["jobName"] += ".small_job"
 
-    with utils.run_job(args.rest, big_job_spec) as big_job:
-        with utils.run_job(args.rest, small_job_spec) as small_job:
-            state = small_job.block_until_state_not_in({"unapproved"})
-            assert state == "queued"
+    with utils.vc_setting(args.rest, args.vc, args.email,
+                          {"scheduling_policy": "FIFO"}):
+        with utils.run_job(args.rest, big_job_spec) as big_job:
+            with utils.run_job(args.rest, small_job_spec) as small_job:
+                state = small_job.block_until_state_not_in({"unapproved"})
+                assert state == "queued"
 
-            state = utils.get_job_status(args.rest, big_job.jid)["jobStatus"]
-            # if vc has user_quota, it will be unapproved
-            assert state in {"queued", "unapproved"}
+                state = utils.get_job_status(args.rest,
+                                             big_job.jid)["jobStatus"]
+                # if vc has user_quota, it will be unapproved
+                assert state in {"queued", "unapproved"}
 
-            expected = "blocked by job with higher priority"
-            for _ in range(50):
-                details = utils.get_job_detail(args.rest, args.email,
-                                               small_job.jid)
+                expected = "blocked by job with higher priority"
+                for _ in range(50):
+                    details = utils.get_job_detail(args.rest, args.email,
+                                                   small_job.jid)
 
-                message = utils.walk_json(details, "jobStatusDetail", 0,
-                                          "message")
-                if expected in message:
-                    break
+                    message = utils.walk_json(details, "jobStatusDetail", 0,
+                                              "message")
+                    if expected in message:
+                        break
 
-                time.sleep(0.5)
-            assert expected in message, "unexpected detail " + str(details)
+                    time.sleep(0.5)
+                assert expected in message, "unexpected detail " + str(details)
 
 
 @utils.case()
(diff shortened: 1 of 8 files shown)


src/utils/JobRestAPIUtils.py
```diff
@@ -944,6 +944,112 @@ def get_vc(username, vc_name):
...
+                    meta, "admin", "job_max_time_second")
+                result["interactive_limit"] = walk_json(meta, "admin",
+                                                        "interactive_limit")
+                return result, 200
+        else:
+            return {
+                "error":
+                    "%s do not have permission to query meta from vc %s" %
+                    (username, vc_name)
+            }, 401
```
Anbang-Hu: Should be 403? 401 usually means "when authentication is required and has failed or has not yet been provided"

src/utils/JobRestAPIUtils.py
```diff
@@ -944,6 +944,112 @@ def get_vc(username, vc_name):
...
+def get_vc_meta(username, vc_name):
+    try:
+        if has_access(username, VC, vc_name, ADMIN):
+            result = {}
+            with DataHandler() as data_handler:
+                meta = json.loads(data_handler.GetVC(vc_name)["metadata"])
+                result["scheduling_policy"] = walk_json(meta,
+                                                        "admin",
+                                                        "scheduling_policy",
+                                                        default="RF")
```
Anbang-Hu, line 956: default should be `FIFO`

src/utils/JobRestAPIUtils.py
```diff
@@ -944,6 +944,112 @@ def get_vc(username, vc_name):
...
+                    logger.info("%s update vc meta for %s to %s", username,
+                                vc_name, json.dumps(meta))
+                    data_handler.UpdateVCMeta(vc_name, json.dumps(meta))
+                    return {"error": None}, 200
+        else:
+            return {
+                "error":
+                    "%s do not have permission to query meta from vc %s" %
+                    (username, vc_name)
+            }, 401
```
Anbang-Hu: I would insist on 403

src/utils/JobRestAPIUtils.py
```diff
@@ -944,6 +944,112 @@ def get_vc(username, vc_name):
...
+                        meta["admin"]["scheduling_policy"] = scheduling_policy
+
+                if "job_max_time_second" in vc_meta:
+                    job_max_time_second = vc_meta.pop("job_max_time_second")
+                    if job_max_time_second is not None and type(
+                            job_max_time_second) != int:
+                        return {
+                            "error":
+                                "job_max_time_second should be int, got %s" %
+                                (job_max_time_second)
```
Anbang-Hu: `"100"` is string, it will print as an integer. Better return the type of the input `job_max_time_second` as well.

src/utils/JobRestAPIUtils.py
```diff
@@ -944,6 +944,112 @@ def get_vc(username, vc_name):
...
+                        }, 400
+                    else:
+                        meta["admin"]["interactive_limit"] = interactive_limit
+
+                if len(vc_meta) > 0:
+                    return {
+                        "error": "unknown key(s) %s" % (str(vc_meta.keys()))
+                    }, 400
+                else:
+                    logger.info("%s update vc meta for %s to %s", username,
```
Anbang-Hu: Do you think it's a good idea to log the vc meta before update as well?

src/ClusterManager/test/test_regular_job.py
```diff
@@ -965,6 +968,45 @@ def test_set_max_time_works(args):
...
+def test_can_not_create_endpoint_if_interactive_limit(args):
+    job_spec = utils.gen_default_job_description("regular", args.email,
+                                                 args.uid, args.vc)
+
+    with utils.vc_setting(args.rest, args.vc, args.email,
+                          {"interactive_limit": 0}):
+        with utils.run_job(args.rest, job_spec) as job:
+            endpoints = utils.create_endpoint(args.rest, args.email, job.jid,
+                                              ["ssh"])
+            assert False, "should not post endpoints successfully"
```
Anbang-Hu, line 1009: Why not checking the output? `assert False` always raises AssertError.

src/ClusterManager/test/test_regular_job.py
```diff
@@ -965,6 +968,45 @@ def test_set_max_time_works(args):
...
+            state = job.block_until_state_not_in({"running"}, timeout=30)
+            assert state == "killed"
+            expected = "running exceed pre-defined 5s"
+
+            details = utils.get_job_detail(args.rest, args.email, job.jid)
+            message = details.get("errorMsg")
+            assert message == expected, "unexpected message " + message
+
+
+#@utils.case(dangerous=True) should have a way to create a normal user
```
Anbang-Hu, line 999: We can create a user `dummy` in DB and use it for testing a regular user.

Anbang-Hu (approved):

src/utils/JobRestAPIUtils.py
xudifsd: Are you sure? FIFO means big job will block later smaller jobs

src/ClusterManager/test/test_regular_job.py
xudifsd: I'm thinking creating a brand new database before doing tests and tear down after it. We might need a framework to create service and let them discover each other on k8s for testing. Disable this test case for now.

src/utils/JobRestAPIUtils.py
Anbang-Hu: I got it wrong. RF should be the default.

PR #1195, opened 2020-06-19
add a init program

xudifsd:
Some benefit of init program:

* reap zombie processes
* capture signal and print out for debug
* capture exit signal and broadcast to user's process, allow graceful exit, also improve exit speed

compiled init program has 2.2MB, and takes 5s to terminate the pod instead of 30s.

Before this, integration test takes 3m56s:
```
real    3m56.608s
user    0m11.417s
sys     0m2.814s
```

After this, takes 2m35s:
```
real    2m35.153s
user    0m10.407s
sys     0m2.847s
```

### DIFF
--- a/src/docker-images/init-container/runtime/init.go
+++ b/src/docker-images/init-container/runtime/init.go
@@ -1,0 +1,88 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"os/signal"
+	"syscall"
+	"time"
+)
+
+func main() {
+	if len(os.Args) == 1 {
+		fmt.Fprintf(os.Stderr, "[init] Usage: %s cmd to run\n", os.Args[0])
+		os.Exit(1)
+	}
+
+	stopCh := make(chan bool, 1)
+	go waitOrphans(stopCh)
+
+	err := run(os.Args[1:]...)
+
+	stopCh <- true
+
+	if exiterr, ok := err.(*exec.ExitError); ok {
+		if status, ok := exiterr.Sys().(syscall.WaitStatus); ok {
+			now := time.Now()
+			fmt.Fprintf(os.Stderr, "[init] %v exit code: %d\n", now.UTC(), status.ExitStatus())
+			os.Exit(status.ExitStatus())
+		} else {
+			now := time.Now()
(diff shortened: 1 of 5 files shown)


src/init-scripts/bootstrap.sh
```diff
@@ -39,15 +39,9 @@ set +e
...
 if [ "$DLTS_ROLE_NAME" = "worker" ];
 then
-    runuser -l ${DLTS_USER_NAME} -c "sleep infinity"
+    exec /usr/local/bin/init runuser -l ${DLTS_USER_NAME} -c "sleep infinity"
 else
     printenv DLTS_LAUNCH_CMD > /dlts-runtime/status/job_command.sh
     chmod +x /dlts-runtime/status/job_command.sh
-    runuser -s /bin/bash -l ${DLTS_USER_NAME} -c /dlts-runtime/status/job_command.sh
-    # Save exit code
-    EXIT_CODE=$?
```
Anbang-Hu, line 48: We don't need `EXIT_CODE` any more?

src/docker-images/init-container/runtime/init.go
```diff
@@ -0,0 +1,88 @@
...
+	cmd := exec.Command(command[0], command[1:]...)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+
+	go func() {
+		for sig := range sigs {
+			if sig != syscall.SIGCHLD {
+				fmt.Fprintf(os.Stderr, "[init] received signal %d\n", sig)
+				syscall.Kill(-cmd.Process.Pid, sig.(syscall.Signal))
```
Anbang-Hu, line 82: Not familiar. Just want to confirm `-` before pid is needed?

Anbang-Hu (approved):

src/docker-images/init-container/runtime/init.go
xudifsd: Yes, this means send signal to that process group. We let Command use a new process group before run it.

src/init-scripts/bootstrap.sh
xudifsd: Yes, because we use `exec`, means no bash code will be executed after it.

xudifsd: The init process will handle this

PR #1203, opened 2020-06-23
Add global team notifications

Gerhut:
Add notification to dashboard yaml config:

```yaml
frontend:
  notifications:
    team_id: the quick brown fox jumps over the lazy dog.
```

And it will display in all pages if user is selecting this team:

image


Anbang-Hu:
This is a good feature. We can extend it to support configurable notification from different clusters and VCs. Might also be a good place to show VC policy?

Anbang-Hu (approved):

PR #1209, opened 2020-06-25
adjust inference params to suit our needs

xudifsd:
[Ref](https://github.com/microsoft/frameworkcontroller/blob/master/doc/user-manual.md#retrypolicy) for Retry policy. [Ref](https://github.com/microsoft/frameworkcontroller/blob/master/doc/user-manual.md#frameworkattemptcompletionpolicy) for  CompletionPolicy.

We want inference recreate pod if it is being preempted by other pod, and do not fail the whole attempt.

### DIFF
--- a/src/ClusterManager/framework.py
+++ b/src/ClusterManager/framework.py
@@ -19,7 +19,7 @@ class Resource(object):
 
 class Role(object):
     def __init__(self, name, task_count, image, envs, resource,
-                 min_failed_task_count, succeeded_task_count):
+                 min_failed_task_count, succeeded_task_count, max_retry_count):
         self.name = name
         self.task_count = task_count
         self.image = image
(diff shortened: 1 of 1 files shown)


src/ClusterManager/framework.py
```diff
@@ -807,10 +808,10 @@ def transform_inference_job(params, cluster_config):
 
     roles = {
         "master":
-            Role("master", 1, image, envs, master_resource, 1, 1),
+            Role("master", 1, image, envs, master_resource, -1, -1, -1),
```
Anbang-Hu, line 811: @leigaoms can you help confirm if this is the desired terminating condition for an inference job. This basically says only when the sole `master` succeeds, the job finishes successfully, otherwise any failure in `master` or `worker` will trigger retry indefinitely.

leigaoms: We need to separate if the pod is preempted or it's really failed. Approve this PR first, and I will add another function to judge the exit code of the process.

leigaoms (approved):

PR #1219, opened 2020-06-30
support preemptable inference jobs

leigaoms:
1. RestAPI: 
    - SubmitJob(): parse "mingpu", "maxgpu" in jobParams, and also be compatible with "resourcegpu", "gpulimit".
    - ScaleJob(): change "resourcegpu" to "mingpu" and "maxgpu". No compatibility issue.
2. JobManager:
   - Add "job_preemptable_resource", "allowed_resource" in job_info. If inference job is scheduling/running, subtract “mingpu” related resource, and keep the job for preempt gpu allocation later.
   - Scheduling logic:
   1) Mark non-preempt training job: job status is "queued".
   2) Mark inference job non-preempt part: job status is "queued". Allocate if all "mingpu" related resource could be satisfied.
   3) Mark preempt training job: job status is "queued/scheduling/running".
   4) Mark inference job preempt part: job status is "queued/scheduling/running". Allocate partial resource if not all "maxgpu" related resource could be satisfied. Suppose CPU/Memory is more than GPU, so allocate GPU first, CPU and memory is in proportion with allocated GPU.

Refinement:
1. fair-sharing: In 4), share remaining GPU among different inference jobs in proportion or on average.
2. Assumption does not apply for CPU cluster, CPU inference job

### DIFF
--- a/src/ClusterManager/job_launcher.py
+++ b/src/ClusterManager/job_launcher.py
@@ -1081,7 +1081,7 @@ class PythonLauncher(Launcher):
             conditionFields = {"jobId": job_object.job_id}
             dataHandler.UpdateJobTextFields(conditionFields, dataFields)
         except Exception as e:
-            logger.error("Submit job failed: %s, %s" % (job, str(e)), exc_info=True)
+            logger.exception("Submit job failed: %s" % job, exc_info=True)
             ret["error"] = str(e)
             retries = dataHandler.AddandGetJobRetries(job["jobId"])
             if retries >= 5:
@@ -1158,7 +1158,7 @@ class PythonLauncher(Launcher):
         name = job_object.job_id + "-1-deployment"
         deployment = self._get_deployment(name=name)
         replicas = deployment.spec.replicas
-        new_replicas = int(job_object.params["resourcegpu"] - job_object.params["mingpu"])
+        new_replicas = int(job_object.params["resourcegpu"]) - int(job_object.params["mingpu"])
         if replicas == new_replicas:
             return
 
--- a/src/ClusterManager/job_launcher.py
+++ b/src/ClusterManager/job_launcher.py
@@ -1081,7 +1081,7 @@ class PythonLauncher(Launcher):
             conditionFields = {"jobId": job_object.job_id}
             dataHandler.UpdateJobTextFields(conditionFields, dataFields)
         except Exception as e:
-            logger.error("Submit job failed: %s" % job, exc_info=True)
+            logger.error("Submit job failed: %s, %s" % (job, str(e)), exc_info=True)
             ret["error"] = str(e)
             retries = dataHandler.AddandGetJobRetries(job["jobId"])
             if retries >= 5:
@@ -1155,10 +1155,10 @@ class PythonLauncher(Launcher):
         if job_object.params["jobtrainingtype"] != "InferenceJob":
             return
 
-        name = job_object.job_id + "-deployment"
+        name = job_object.job_id + "-1-deployment"
         deployment = self._get_deployment(name=name)
         replicas = deployment.spec.replicas
-        new_replicas = int(job_object.params["resourcegpu"])
+        new_replicas = int(job_object.params["resourcegpu"] - job_object.params["mingpu"])
         if replicas == new_replicas:
             return
 
--- a/src/ClusterManager/job_manager.py
+++ b/src/ClusterManager/job_manager.py
@@ -556,6 +556,7 @@ def get_jobs_info(jobs, cluster_schedulable, vc_schedulables):
         if job_status in ["queued", "scheduling", "running"]:
             allowed = False
             allowed_resource = None
+
             job_params = json.loads(base64decode(job["jobParams"]))
             preemption_allowed = job_params.get("preemptionAllowed", False)
             job_id = job_params["jobId"]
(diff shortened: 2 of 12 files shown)


src/ClusterManager/job_manager.py
```diff
@@ -720,20 +735,119 @@ def mark_schedulable_preemptable_jobs(jobs_info, cluster_schedulable):
...
+        # inference job which non-preemptable part is not allowed or has no preemptable part
+        if job_info["jobtrainingtype"] != "InferenceJob" or \
+            job_info["allowed"] is False or \
+            job_info["job_preemptable_resource"] is None:
+            continue
+
+        job_preemptable_resource = job_info["job_preemptable_resource"]
+        job_id = job_info["jobId"]
+
+        if cluster_schedulable != empty_resource:
```
xudifsd: Why we need this check?

xudifsd (commented):
Seems inference job do not read preemptionAllowed and split one inference job as preemptible and non-preemptible part, never preempt non-preemptible part. Correct?

src/ClusterManager/job_manager.py
leigaoms: Because we want to allocate resource as long as cluster resource is not empty, no matter cluster resource is sufficient or not.  If cluster_resource is sufficient, we will allocate maxgpu. If not, we will allocate cluster's remaining gpu.

leigaoms (commented):
> Seems inference job do not read preemptionAllowed and split one inference job as preemptible and non-preemptible part, never preempt non-preemptible part. Correct?

Yes. When submitting inference job, set "preemptionAllowed" to True.

src/ClusterManager/job_manager.py
xudifsd: I think this will never be False. For example some nodes have some free CPU.

xudifsd:
Have you tested these, or is behaviour will like these:

* inference job being preempted by training jobs, training jobs squeeze resource used by inference to min_gpu and no more regular job can be scheduled(I'm not sure what will happen if at this time inference job owner increase min_gpu)
* training job can preempt other training jobs with preemption enabled but not inference job with min_gpu allocated
* inference job can be allocate more if resource is available
* inference job can not allocate more if other training jobs are also waiting resource

src/ClusterManager/job_manager.py
Anbang-Hu: `cluster_schedulable` will hardly be empty. E.g.
```
"gpu": {"Standard_ND24rs": 0},
"cpu": {"Standard_ND24rs": 100},
"memory": {"Standard_ND24rs": "100Gi"},
```
In this case, the `schedulable_resource` here becomes 0. Then the logging at 808 doesn't sound appropriate, because no preemptable part has been admitted for scheduling.

leigaoms:
Have you tested these, or is behaviour will like these:

- inference job being preempted by training jobs, training jobs squeeze resource used by inference to min_gpu and no more regular job can be scheduled(I'm not sure what will happen if at this time inference job owner increase min_gpu)
**Lei:** Good point! I have added a constraint: User cannot change min_gpu if job is scheduling/running. If they need to change min_gpu, pause the job first to release all GPUs, then resume it. For already scheduling/running inference job, min_gpu cannot be preempted.

- training job can preempt other training jobs with preemption enabled but not inference job with min_gpu allocated
**Lei:** Yes

- inference job can be allocate more if resource is available
**Lei:** Yes, up to min(max_gpu, remaining cluster gpu)

- inference job can not allocate more if other training jobs are also waiting resource
**Lei:** Yes. training job is scheduled ahead of inference job

src/ClusterManager/job_manager.py
leigaoms: Got it, I will resolve it

xudifsd (approved):

src/RestAPI/dlwsrestapi.py
```diff
@@ -314,15 +314,17 @@ def __init__(self):
         self.get_parser = reqparse.RequestParser()
         self.get_parser.add_argument("jobId", required=True)
         self.get_parser.add_argument("userName", required=True)
-        self.get_parser.add_argument("resourcegpu", required=True)
+        self.get_parser.add_argument("mingpu", required=True)
+        self.get_parser.add_argument("maxgpu", required=True)
 
     def get(self):
```
Gerhut: Shall we use `POST` to do modifications?

leigaoms: ok

src/ClusterManager/job_launcher.py
```diff
@@ -1081,7 +1081,7 @@ def submit_job_impl(self, job):
             conditionFields = {"jobId": job_object.job_id}
             dataHandler.UpdateJobTextFields(conditionFields, dataFields)
         except Exception as e:
-            logger.error("Submit job failed: %s" % job, exc_info=True)
+            logger.error("Submit job failed: %s, %s" % (job, str(e)), exc_info=True)
```
xudifsd: It's better to use logger.exception, as it can print out stack trace.

src/ClusterManager/job_launcher.py
```diff
@@ -1155,10 +1155,10 @@ def scale_job(self, job):
         if job_object.params["jobtrainingtype"] != "InferenceJob":
             return
 
-        name = job_object.job_id + "-deployment"
+        name = job_object.job_id + "-1-deployment"
         deployment = self._get_deployment(name=name)
         replicas = deployment.spec.replicas
-        new_replicas = int(job_object.params["resourcegpu"])
+        new_replicas = int(job_object.params["resourcegpu"] - job_object.params["mingpu"])
```
xudifsd: I think this should be casting two params to int before doing minus.

src/ClusterManager/pod_template.py
```diff
@@ -226,8 +226,10 @@ def generate_pods(self, job):
 
         deployment_params = copy.deepcopy(params)
 
-        deployment_params["deployment_replicas"] = params["resourcegpu"]
+        deployment_params["deployment_replicas"] = params["mingpu"]
         deployment_params["LaunchCMD"] = params["cmd"]
+        deployment_params["preemptionAllowed"] = False
+        deployment_params["preemptable"] = "0"
```
xudifsd: Please give it a reasonable name maybe `deploymentIndex`, preemptable makes me think the value should be True or False.

Also, please add comment somewhere mentioning why we need two deployments.

PR #1250, opened 2020-07-14
ctl.py partially supports onpremise cluster maintenance

YinYangOfDao:
update the pipeline tool a little bit to support onpremise machine management


xudifsd (approved):

xudifsd:
Can this work on Lab's config?

YinYangOfDao:
> Can this work on Lab's config?

this is for Lab's config. Currently there's no other type of on-premise clusters, so this is exactly for lab machines. This PR should be used with converted config files that fit in cloud-init pipeline.

PR #1269, opened 2020-07-23
cpu inference

leigaoms:
Add CPUInferenceJob in job submission, and scheduling


xudifsd (approved):

xudifsd:
Can we add a different parameter to InferenceJob instead of adding a CPUInferenceJob? I really do not want to add another job type.

PR #1283, opened 2020-08-06
add aml link config and add jobId to email template

zhangqilike2015:
add aml link config and add jobId to email template

### DIFF
--- a/src/dashboard/src/pages/Jobs/index.tsx
+++ b/src/dashboard/src/pages/Jobs/index.tsx
@@ -1,12 +1,13 @@
 import * as React from 'react'
-import { useEffect, useState } from 'react'
 import {
+  useEffect, useState,
   ChangeEvent,
   FunctionComponent,
   useCallback,
   useContext,
   useMemo
 } from 'react'
+
 import {
   useHistory,
   useLocation,
@@ -88,12 +89,12 @@ const Jobs: FunctionComponent = () => {
   const history = useHistory()
   const { hash } = useLocation()
   const { clusterId } = useParams<RouteParams>()
-  const [amlUrl, setAmlUrl ] = useState('');
+  const [amlUrl, setAmlUrl] = useState('')
   const fetchAmlUrl = '/api/clusters'
   const requestAmlUrl = useFetch(fetchAmlUrl)
   const fetchAml = async () => {
-    for (var i in clusters) {
-      const { amlPortal } = await requestAmlUrl.get(`/${clusters[i].id}`)
+    for (const item of clusters) {
+      const { amlPortal } = await requestAmlUrl.get(`/${item.id}`)
       if (amlPortal != null && amlPortal != '') {
         setAmlUrl(amlPortal)
       }
@@ -101,7 +102,7 @@ const Jobs: FunctionComponent = () => {
   }
   useEffect(() => {
     fetchAml()
-  }, [])
+  }, [fetchAml])
 
   const cluster = useMemo(() => {
     return clusters.filter(cluster => cluster.id === clusterId)[0]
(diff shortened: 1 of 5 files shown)


src/dashboard/src/pages/Jobs/index.tsx
```diff
@@ -82,6 +88,20 @@ const Jobs: FunctionComponent = () => {
...
   const { hash } = useLocation()
   const { clusterId } = useParams<RouteParams>()
+  const [amlUrl, setAmlUrl ] = useState('');
+  const fetchAmlUrl = '/api/clusters'
+  const requestAmlUrl = useFetch(fetchAmlUrl)
+  const fetchAml = async () => {
+    for (var i in clusters) {
+      const { amlPortal } = await requestAmlUrl.get(`/${clusters[i].id}`)
+      if (amlPortal != null && amlPortal != '') {
+        setAmlUrl(amlPortal)
```
Gerhut, line 99: It seems we are fetching `amlPortal` of all clusters and overwrite them one by one

src/dashboard/src/pages/Jobs/index.tsx
```diff
@@ -99,6 +119,13 @@ const Jobs: FunctionComponent = () => {
     <Container fixed maxWidth="xl">
       <Toolbar disableGutters>
         <FormControl fullWidth>
+          { amlUrl != '' ?
```
Gerhut: I got what you want, just put a `useFetch` with full url and make the `clusterId` as its dependency. and then use a `useMemo` to map the cluster config to the actual `amlUrl`

src/dashboard/src/pages/Jobs/index.tsx
```diff
@@ -14,17 +15,22 @@ import {
...
+  Tooltip,
   Toolbar
 } from '@material-ui/core'
 import SwipeableViews from 'react-swipeable-views'
 
 import ClustersContext from '../../contexts/Clusters'
 import ClusterSelector from '../../components/ClusterSelector'
 
+import useFetch from 'use-http'
+import { Info } from '@material-ui/icons'
```
Gerhut: Put @material-ui/icons at the bottom of @material-ui/core to prevent repeat imports

src/dashboard/src/pages/Jobs/index.tsx
```diff
@@ -82,6 +88,20 @@ const Jobs: FunctionComponent = () => {
   const history = useHistory()
   const { hash } = useLocation()
   const { clusterId } = useParams<RouteParams>()
+  const [amlUrl, setAmlUrl ] = useState('');
```
Gerhut: Use `undefined` as some pending value instead of empty string.

Gerhut (changes requested):

src/dashboard/src/pages/Submission/DataJob.tsx
```diff
@@ -8,10 +8,13 @@ import {
   CardActions,
   Grid,
   Container,
+  Tooltip,
   TextField,
   Button,
   Divider, Dialog, DialogTitle, DialogContent, DialogContentText, DialogActions
 } from '@material-ui/core'
+import {Link as AmlLink}  from '@material-ui/core'
```
Gerhut: Why rename `Link` here?

zhangqilike2015: Same name with component in Row 24.
import { Link } from 'react-router-dom'

Gerhut: I see, there is a best practice for you to follow

https://github.com/microsoft/DLWorkspace/blob/master/src/dashboard/src/pages/Jobs/JobsTable/columns/useNameId.tsx#L7-L8
https://github.com/microsoft/DLWorkspace/blob/master/src/dashboard/src/pages/Jobs/JobsTable/columns/useNameId.tsx#L43-L45

PR #1288, opened 2020-08-07
Dontfailonswap

YinYangOfDao:
This --fail-swap-on parameter basically makes the Kubelet fail to start if swap is enabled on the node.


Anbang-Hu (changes requested):
Changing 207 files does not look correct. Please only cherry-pick the change instead of PR your entire dev branch into v1.7.

PR #1307, opened 2020-08-26
update mt for exemption

SummerX:
image
image
image


src/dashboard/server/api/validator/index.js
```diff
@@ -7,6 +7,7 @@ validator.addSchema(require('./endpoints.schema'), 'endpoints')
 validator.addSchema(require('./job.schema'), 'job')
 validator.addSchema(require('./template.schema'), 'template')
 validator.addSchema(require('./priority.schema'), 'priority')
+validator.addSchema(require('./exemption.schema'), 'isExempted')
```
Gerhut, line 10: We'd better use same names in resource name and field name instead of `exemption` and `isExempted`, at least, the second parameter stands for the schema name of the exemption resource, should be `exemption`

src/dashboard/src/hooks/useActions.ts
```diff
@@ -28,6 +28,16 @@ const KILLABLE_STATUSES = [
...
 ];
 
+const EXEMPTABLE_STATUSES = [
+  'unapproved',
+  'queued',
+  'scheduling',
+  'running',
+  'pausing',
+  'paused',
+  'failed'
```
Gerhut, line 38: Why is failed job allowed to set exemption?

src/dashboard/src/hooks/useActions.ts
```diff
@@ -173,7 +210,19 @@ ${givenName} ${familyName}
     }
   }, { position: 'row' }), [onKill]);
 
-  return { support, approve, pause, resume, kill };
+  const exemption = useCallback(Object.assign((job: any): Action<any> => {
+    const hidden = EXEMPTABLE_STATUSES.indexOf(job['jobStatus']) === -1;
+    const isExempted = job['jobParams'].hasOwnProperty('policy_exemptions') && Array.isArray(job['jobParams']['policy_exemptions']) && job['jobParams']['policy_exemptions'].length > 0 ;
```
Gerhut, line 215: Using prototype methods may introduce security concerns, use lodash.get instead.

Gerhut (changes requested):

PR #1310, opened 2020-08-28
update wording for exemption & fix the first notification not clickable.

SummerX:


MaggieHust (approved):

src/dashboard/src/Layout/Content.tsx
```diff
@@ -34,7 +35,7 @@ const Content: FunctionComponent = ({ children }) => {
   return (
     <Box flex={1} py={3} className={styles.root}>
```
Gerhut:
```suggestion
    <Box flex={1} py={3} pt={0} className={styles.root}>
```

instead of the top change.

Gerhut:
Seems confliction

PR #1314, opened 2020-08-31
Add a new api which can get gpu_pernode and gputype

lupen0120:
We want to have an Aether REST API, given team/cluster/userName, it could return #gpusPerNode and gpuType.

As the #gpusPerNode and gpuType would be quite different in different team/clusters, for example the resourceGpu in Azure-WestUS2-V100-16GB-LowPriority is 4; Azure-NorthCentralUS-V100 is 8. For our users, resourceGpu is the No.1 most frequently wrong parameter. When users switch team/cluster, lots of them don’t know they need to change resourceGpu as well.

for example：https://dltshub.redmond.corp.microsoft.com//api/teams/{$0}/clusters/{$1}/metadata?email={$2}
the result is : {type: {$0}, per_node: {$1}}


src/dashboard/server/api/router.js
```diff
@@ -25,6 +25,9 @@ router.get('/teams/:teamId/clusters/:clusterId',
 router.get('/clusters/:clusterId',
   require('./middlewares/user')(),
   require('./controllers/cluster'))
+router.get('/teams/:teamId/clusters/:clusterId/metadata',
+  require('./middlewares/user')(),
+  require('./controllers/team/metadata'))
```
Gerhut: It's better to put the controller to
```suggestion
  require('./controllers/team/cluster/metadata'))
```

Gerhut:
Could you change the base branch from master to v1.7?

PR #1315, opened 2020-08-31
Lupen/dlts metadata v1.7

lupen0120:
We want to have an Aether REST API, given team/cluster/userName, it could return #gpusPerNode and gpuType.

As the #gpusPerNode and gpuType would be quite different in different team/clusters, for example the resourceGpu in Azure-WestUS2-V100-16GB-LowPriority is 4; Azure-NorthCentralUS-V100 is 8. For our users, resourceGpu is the No.1 most frequently wrong parameter. When users switch team/cluster, lots of them don’t know they need to change resourceGpu as well.

for example：https://dltshub.redmond.corp.microsoft.com//api/teams/{$0}/clusters/{$1}/metadata?email={$2}
the result is : {type: {$0}, per_node: {$1}}


src/dashboard/server/api/controllers/team/cluster/metadata.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * @typedef {Object} State
+ * @property {import('../../../services/cluster')} cluster
+ */
+
+/** @type {import('koa').Middleware<State>} */
+module.exports = async context => {
+    const { cluster } = context.state
```
Gerhut: Incorrect indention here, `npm run lint:backend -- --fix` will help you

